PR #8, opened 2023-03-19
docs: clarify premium price id in `constants.ts`

### LINKED ISSUE #7: Not setting own premium plan price id results in a 500 error at stripe
After setting up a new project following the `README` i found that not setting the `STRIPE_PREMIUM_PLAN_PRICE_ID` in `constants.ts` to my created premium plan resulting in the following error:
```
Server error

500 internal error: No such price: 'price_1MPiEkD6QJts4RaYcp1SevPe'
```

The price id mentioned is the one that is preconfigured in the `constants.ts`. Maybe updating the `Customizing Deno SaaSKit` section in `README` to give a heads up might help preventing this error.

However setting the mentioned `STRIPE_PREMIUM_PLAN_PRICE_ID` variable to an empty string `""` in `constants.ts` will result in a differend error:
```
500 internal error: You passed an empty string for 'line_items[0][price]'. We assume empty values are an attempt to unset a parameter; however 'line_items[0][price]' cannot be unset. You should remove 'line_items[0][price]' from your request or supply a non-empty value.
```

So first I would suggest updating the `README` to make sure that a new starter can find the solution there.

niklasmtj:
When not setting `STRIPE_PREMIUM_PLAN_PRICE_ID` to one's own id Stripe will not be able to find the product. This will fail with a 500 error.

Fixes #7

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,7 +161,13 @@ export const STRIPE_PREMIUM_PLAN_PRICE_ID = "price_1MPiEkD6QJts4RaYcp1SevPe";
 export const FREE_PLAN_TODOS_LIMIT = 5;
 ```
 
-The Stripe premium plan price id is the "API ID" of your Stripe product.
+The Stripe premium plan price id is the "API ID" of your Stripe product. If you
+do not set it to your own premium plan price id, Stripe will not be able to find
+the set product plan when it tries to upgrade the subscription from the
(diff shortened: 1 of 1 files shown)


iuioiua:
Good spot and fix! Thank you very much. LGMT.

PR #10, opened 2023-03-21
feat: basic theme customization

### LINKED ISSUE #9: Basic theme customization
Todo: explain the basics of theme customization by example and explanation.

iuioiua:
This PR defines `primary` and `secondary` colors within `twind.config.js` for the site-wide theme, providing a foundation for further theme customization. A corresponding section within the README was also added. This can be expanded upon in the future if need be.

Closes #9.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,7 +224,7 @@ TODO.
 
 You can customize theme options such as spacing, color, etc. By default, Deno
 SaaSKit comes with `primary` and `secondary` colors predefined within
-`twind.config.ts`. Change these values to match your desired color-scheme.
+`twind.config.ts`. Change these values to match your desired color scheme.
 
 ## Building a Modern SaaS Stack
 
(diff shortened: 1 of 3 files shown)


iuioiua:
@hashrock, do you know how to get these preview deployments to read the environment secrets I've defined on the settings page?

kt3k:
Deno Deploy still can read env variable in preview deployment, but if the env var is set in the project, then preview deployments are only created for `push` event of the origin repository, and it doesn't create previews for PRs from forks (to prevent stealing env vars from fork PRs)

iuioiua:
Hm... I see. Are there any workarounds? I want to be able to preview deployments before merging PRs. One alternative that comes to mind is to change any code requiring environmental variables so that it's no longer required but skipped in some way. Simpler, repo-level options would be preferred.

kt3k:
> Hm... I see. Are there any workarounds?

I think the simple workaround is to move your branch to this repo and directly push to there (that causes `push` event and preview should be available even with env vars).

PR #16, opened 2023-03-26
refactor: webhooks-based approach

iuioiua:
Previously, newly created OAuth users weren't registered as Stripe customers, so their subscriptions couldn't be managed.

This change adds a webhook endpoint for Stripe. When a user's subscription is renewed or cancelled, `/api/subscription` is called, making a corresponding change to the status of the subscription in the `customers` table in Supabase.

This PR also includes other minor cleanups, fixes and tweaks.

#19 is a prerequisite for this PR.


iuioiua:
@thorwebdev, is there any chance you'd be able to have a quick look over this? Is this an improvement in design?

.example.env
```diff
@@ -1,4 +1,8 @@
 SUPABASE_ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0
 SUPABASE_URL=http://localhost:54321
+SUPABSE_SERVICE_KEY=xxx
 
-STRIPE_SECRET_KEY=sk_test_xxx
\ No newline at end of file
+STRIPE_SECRET_KEY=sk_test_xxx
+STRIPE_SIGNING_SECRET=xxx
```
thorwebdev, line 6: I'd call it `STRIPE_WEBHOOK_SECRET`

README.md
```diff
@@ -106,6 +109,74 @@ You can also keep the column `created_at` if you'd like.
...
+- Go to `Database` > `Triggers`
+- Click `Create a new trigger` and enter the following values:
+  - Name of trigger = `new_customer`
+  - Table = `users auth`
+  - Events = `Insert`
+  - Trigger type = `After the event`
+  - Orientation = `Row`
+  - Function to trigger = `create_customer`
+
+### Automate Stripe subscription updates via Supabase
```
thorwebdev, line 155: Using Database Webhooks is definitely a good way to keep things in sync in realtime, however it might not be required.

What I would do instead is create a method `createOrRetrieveCustomer` that takes in the user id and returns the stripe customer id if the customer already exists in Stripe, or creates a new stripe customer, stores the stripe customer id in the database, and then returns the stripe customer id.

You can see an implementation of that approach here: https://github.com/vercel/nextjs-subscription-payments/blob/main/utils/supabase-admin.ts#L52-L83

iuioiua: Excellent suggestion! I've gone full circle and reverted webhooks for functionality relating to the `customers` table, and now it's looking much simpler. Thank you.

iuioiua:
Changes made. PTAL, @thorwebdev.

iuioiua:
I think there'd be fewer moving parts if `stripe_customer_id` and `is_subscribed` were stored in the user's metadata object rather than a `customers` table. WDYT?

iuioiua:
I'll go ahead and merge for now. We can evolve on these ideas later.

routes/dashboard/_middleware.ts
```diff
@@ -26,18 +26,10 @@ export async function handler(
...
-      .single()
-      .throwOnError();
 
-    ctx.state.subscription = {
-      stripeCustomerId: data!.stripe_customer_id,
-      isSubscribed: data!.is_subscribed,
-    };
+    ctx.state.session = session;
+    ctx.state.supabaseClient = supabaseClient;
+    ctx.state.customer = await createOrGetCustomer(supabaseClient, stripe);
```
thorwebdev, line 32: This would make a DB request every time the middleware is executed, no? 

I'd say we shouldn't call `createOrGetCustomer` in the middleware, but only actually call it when we need to access the stripe customer id. Wdyt?

thorwebdev:
Sorry for the delay. Yes, LGTM 👍 just added one note above regarding using `createOrGetCustomer` in the middleware.

routes/dashboard/_middleware.ts
iuioiua: Good point. I've turned it into a helper to be called whenever needed now in #28.

PR #19, opened 2023-03-28
chore: some supabase and stripe suggestions.

thorwebdev:
First off, this is looking awesome 🙌 well done y'all! 💚

Here some suggestions that I think could be helpful:

- Default to local supabase development
  - add initial migration
  - add local env vars
- Update policy to work for INSERTs also (WITH CHECK)
- Update `user_id` to be filled automatically via `auth.uid()` method 
- Add `formatAmountForDisplay` method that uses the `Intl.NumberFormat` api

Let me know what you think :)


.example.env
```diff
@@ -1,4 +1,4 @@
-SUPABASE_ANON_KEY=xxx
-SUPABASE_URL=https://xxx.supabase.co
+SUPABASE_ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0
```
iuioiua, line 1: Where'd this value come from? Does it work on local Supabase instances?

routes/dashboard/api/todo.ts
```diff
@@ -12,9 +12,10 @@ export const handler: Handlers = {
 
       return Response.json(null, { status: 201 });
     } catch (error) {
+      console.log({ error });
       const status = error instanceof AuthError ? 401 : 400;
 
-      return new Response(error, { status });
+      return new Response(error.message, { status });
```
iuioiua, line 18: Nice. Since this is an API, would it be better to return the error as JSON via `Response.json(error, { status })` instead? WDYT?

supabase/migrations/20230328042936_init.sql
iuioiua, line 1: Did you generate this using `supabase migrate ...`? If so, very nice! I've been looking for a way to quickly set up the database for new users in #16.

.example.env
thorwebdev: Yes correct, this is the local development credentials. It's being outputted when running `supabase start` or `supabase status` (when local stack is running already)

routes/dashboard/api/todo.ts
thorwebdev: No strong feelings on this 👍

supabase/migrations/20230328042936_init.sql
thorwebdev: Yes, correct.

I created the table as per your instructions and then ran `supabase db diff --schema public` to get the changes in SQL form and then added that to the migration file. See docs here: https://supabase.com/docs/guides/cli/local-development#diffing-changes

utils/stripe.ts
```diff
@@ -5,3 +5,28 @@ export const stripe = new Stripe(Deno.env.get("STRIPE_SECRET_KEY")!, {
...
 });
+
+const isClient = typeof window === "object";
+
+export function formatAmountForDisplay(
+  amount: number,
+  currency: string,
+): string {
+  const numberFormat = new Intl.NumberFormat(
+    isClient ? navigator.language : "en-US",
```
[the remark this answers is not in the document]
thorwebdev: This should normally be called on the client as to take the language of the user's browser. but not sure what the best way to get the user's browser language is in deno fresh.

routes/dashboard/api/todo.ts
iuioiua: On 2nd thought, this is fine.

iuioiua (approved):

iuioiua:
Ok, fresh doesn't yet support a `deno.jsonc` file (issue [here](https://github.com/denoland/fresh/issues/1060)). I've had to revert the rename. Also, I've omitted the `start:co` task. I feel as though it may confuse some people using the repo.

I've left a reminder as #23 to clean up the `--cached-only` flag once the upstream issue is fixed.

Either way, it LGMT! Thank you very much for your help, @thorwebdev! Much appreciated!

PR #32, opened 2023-04-02
fix: `SUPABASE_SERVICE_KEY` missing letter

niklasmtj:
My guess was that there was a missing letter in `SUPABASE`.


iuioiua:
Good catch! Thank you 🙏🏾

PR #35, opened 2023-04-03
feat: OAuth

iuioiua:
This change adds OAuth functionality for authentication. So far, GitHub is used as an example.

I'm aware that this approach is suboptimal compared to the incoming PKCE approach but would rather enable OAuth sooner rather than later. We'll switch to PKCE once available.

Towards #13.

@thorwebdev and @kt3k , WDYT?


thorwebdev:
Thanks @iuioiua , and sorry for the delay on PCKE. We needed to fix some [issues in gotrue-js](https://github.com/supabase/gotrue-js/releases/tag/v2.18.1) but those have now been released. Hoping we'll land PCKE support in the auth helpers by end of week!

iuioiua:
> Thanks @iuioiua , and sorry for the delay on PCKE. We needed to fix some [issues in gotrue-js](https://github.com/supabase/gotrue-js/releases/tag/v2.18.1) but those have now been released. Hoping we'll land PCKE support in the auth helpers by end of week!

Not a problem! One of the reasons I made this PR was so that we can directly see the improvement once PKCE has landed. It wasn't out of impatience haha

iuioiua:
@thorwebdev, does the implementation look sound to you? It was taken from https://github.com/thorwebdev/everydaylytics.

routes/signup.tsx
```diff
@@ -18,6 +20,7 @@ export default function SignupPage(props: PageProps) {
           {errorMessage === "User already registered" && (
             <Notice>{errorMessage}</Notice>
           )}
+          <OAuthLoginButton provider="github" icon={Icons.GitHub} />
```
kt3k, line 23: I think the button label needs to be more flexible to support other providers.

How about the interface like the below?

```js
<OAuthLoginButton provider="github">
  <Icons.GitHub class="inline mr-2" />
  Log in with GitHub
</OAuthLoginButton>
```

routes/api/oauth.ts
```diff
@@ -0,0 +1,29 @@
...
+import { assert } from "std/testing/asserts.ts";
+import { createSupabaseClient } from "@/utils/supabase.ts";
+import type { Provider } from "@supabase/supabase-js";
+
+export const handler: Handlers = {
+  async POST(request) {
+    const form = await request.formData();
+    const provider = form.get("provider");
+
+    assert(typeof provider === "string");
```
kt3k, line 11: I think this should return `400` response instead of throwing

kt3k:
Other part looks good to me.

routes/signup.tsx
iuioiua: `OAuthLoginButton` does support other providers. Is there something about it that appears that it might not?

routes/api/oauth.ts
iuioiua: Done.

routes/signup.tsx
iuioiua: On 2nd thought, you're right. `OAuthLoginButton` did too much. I've refactored it.

iuioiua:
Changes made:
1. Login/signup pages look prettier.
2. Changed `/login-success` route file to `/login/success` and `/login` to `/login/index.tsx`.
3. If `provider` isn't a string, return HTTP code 400.

PR #39, opened 2023-04-05
image optimization

rroblf01:
The sum of the svg and png was 711.67kb, but if we apply a slight compression (without losing image quality), it is 177.81kb. That is, we are reducing the weight of these images by 75.01%, and therefore, the page will be lighter.


iuioiua:
LGMT. Thank you, @rroblf01!

iuioiua (approved):

PR #43, opened 2023-04-05
chore: add customer table migration.

thorwebdev:
Adding the local migration for the customer table.


iuioiua (approved):
LGTM! Thank you, @thorwebdev?

PR #50, opened 2023-04-06
chore: add basic meta tags to blog

### LINKED ISSUE #31: feature: add meta tags to blog
To enable OpenGraph descriptions, pictures etc. we should also add the most important meta tags to the blog landing and post pages. What do you think?

niklasmtj:
This addes the basic meta tags with already available content. Beyond that we will still need to figure out how we would like to proceed with images etc.

fixes #31


routes/blog/[slug].tsx
```diff
@@ -33,6 +34,15 @@ export default function PostPage(props: PageProps<Post>) {
...
         <title>{post.title}</title>
+        <meta property="og:type" content="website" />
+        <meta property="og:site_name" content={SITE_NAME} />
+        <meta property="og:locale" content="en" />
+        <meta property="og:title" content={post.title} />
+        <meta property="og:description" content={post.summary} />
+        <meta property="og:url" content={props.url.origin} />
+        <meta name="twitter:title" content={post.title} />
+        <meta name="twitter:description" content={post.summary} />
+        <meta name="description" content={post.summary} />
```
Comment (iuioiua) [inline routes/blog/[slug].tsx:45]: Are we able to add `<meta itemprop="name"...>` and `<meta itemprop="description"... >`?

routes/blog/index.tsx
```diff
@@ -36,6 +37,12 @@ export default function BlogIndexPage(props: PageProps<Post[]>) {
     <>
       <Head>
         <title>Blog</title>
+        <meta property="og:type" content="website" />
+        <meta property="og:site_name" content={SITE_NAME} />
+        <meta property="og:locale" content="en" />
+        <meta property="og:title" content="Blog" />
+        <meta property="og:url" content={props.url.origin} />
+        <meta name="twitter:title" content="Blog" />
```
iuioiua, line 45: Can we match the number of tags as in `/blog/[slug].tsx`? We can have something like "This is the blog for Deno SaaSKit" for the description.

iuioiua (changes requested):
Thank you! LGTM, except for a couple of suggested additions.

routes/blog/index.tsx
niklasmtj: Sure, added it in the new PR.

routes/blog/[slug].tsx
Comment (niklasmtj) [reply routes/blog/[slug].tsx:45]: Of course, I added it to `/blog/[slug].tsx` and `/blog/index.tsx` respectively

iuioiua (approved):
LGTM! Thank you, @niklasmtj.

PR #66, opened 2023-04-11
feat: meta tag component

### LINKED ISSUE #55: Feature request: `<Meta />` component
**Is your feature request related to a problem? Please describe.**
The blog, including the index and individual posts, has meta tags, which are a strong base for future meta tags support but are manually written.

**Describe the solution you'd like**
A `<Meta />` component that can be reused anywhere on the website.

**Additional context**
Spec:
- [ ] `<Meta />` component with reasonable defaults and properties such as `title`, `description`, etc.
- [ ] Comments sectioning meta tags by platform. For example, see [this](https://www.heymeta.com/url/www.youtube.com/watch?v=iuJDhFRDx9M) and click **Generate**.
- [ ] Add the new component wherever applicable on the website.

niklasmtj:
This fixes #55. 

I added the newly created `Meta` component also to the `Head` component which is already used on a couple of pages. 

Happy to about any feedback about the component.

### DIFF
--- a/components/Meta.tsx
+++ b/components/Meta.tsx
@@ -1,0 +1,75 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import { SITE_DESCRIPTION, SITE_NAME } from "@/constants.ts";
+
+interface MetaProps {
+  title?: string;
+  description?: string;
+  imageUrl?: string;
+  url?: string;
+}
+
+export default function Meta(props: MetaProps) {
+  const { title, description, imageUrl, url } = props;
+
+  return (
+    <>
+      {/* HTML Meta Tags */}
+      <title>{title || SITE_NAME}</title>
+      <meta
+        name="description"
+        content={description || SITE_DESCRIPTION}
+      />
+      <meta property="og:type" content="website" />
+      <meta property="og:site_name" content={title || SITE_NAME} />
+      <meta property="og:locale" content="en" />
+
+      {/* Google / Search Engine Tags */}
+      <meta
+        itemProp="name"
+        content={title || SITE_NAME}
+      />
+      <meta
+        itemProp="description"
+        content={description || SITE_DESCRIPTION}
+      />
+      <meta
+        itemProp="image"
+        content={imageUrl}
+      />
+
+      {/* Facebook Meta Tags */}
+      <meta
+        property="og:url"
+        content={url}
+      />
+      <meta property="og:type" content="website" />
+      <meta
+        property="og:title"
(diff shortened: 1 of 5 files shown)


components/Meta.tsx
```diff
@@ -0,0 +1,75 @@
...
+    <>
+      {/* HTML Meta Tags */}
+      <title>{title || SITE_NAME}</title>
+      <meta
+        name="description"
+        content={description || SITE_DESCRIPTION}
+      />
+      <meta property="og:type" content="website" />
+      <meta property="og:site_name" content={title || SITE_NAME} />
+      <meta property="og:locale" content="en" />
```
iuioiua, line 24: I believe these 3 lines should be moved down to the Facebook Meta Tags section, and `og:site_name` should just have a value of `SITE_NAME`.

components/Meta.tsx
```diff
@@ -0,0 +1,75 @@
...
+        content={title || SITE_NAME}
+      />
+      <meta
+        property="og:description"
+        content={description || SITE_DESCRIPTION}
+      />
+      <meta
+        property="og:image"
+        content={imageUrl}
+      />
```
iuioiua, line 57: `imageUrl` is optional. I think it's best to only generate this and other `imageUrl`-containing elements, with a conditional that checks for the existence of `imageUrl` first. E.g. `{imageUrl && <Element>}`.

This behaviour should also occur for other properties of `MetaProps` that lack a default value.

components/Meta.tsx
```diff
@@ -0,0 +1,75 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import { SITE_DESCRIPTION, SITE_NAME } from "@/constants.ts";
+
+interface MetaProps {
+  title?: string;
+  description?: string;
+  imageUrl?: string;
+  url?: string;
```
iuioiua, line 8: Are you able to change this to `href`? It might be confusing since the value passed to this parameter is `URL.href`.

iuioiua (changes requested):
Looking good, except for a couple of things. Can you also add somewhere in the README that these meta tags are a feature? A single line should do.

Note: we'll have to validate these meta tags once this PR is pushed to the public website.

components/Meta.tsx
niklasmtj: That's what I also though about. Did not know how Twitter etc. reacts when there is a tag without a value. Definitely makes sense to render these conditionally.

iuioiua (approved):
I love this! Thank you, @niklasmtj.

PR #67, opened 2023-04-12
refactor: replace components with style strings

iuioiua:
Previously, the `Notice`, `Input`, `Button` and `LinkButton` components provided consistent styles. However, these components provided no additional functionality beyond styling.

This change simplifies style consistency by using strings to define their styling instead, removing the need for these 4 components.


kt3k (approved):
Probably it'll be easier to understand the DOM structure by this change. LGTM

PR #68, opened 2023-04-12
feat/chore: type-checking in CI and update dependencies

iuioiua:
This change adds type-checking as a step in CI. This is an improvement in correctness and consistency in types across the codebase.


kt3k (approved):
LGTM

PR #93, opened 2023-04-16
feat: `deno task init:stripe` and further documentation improvements

iuioiua:
Previously, a bit of manual work was required to set up Stripe. This change adds the init:stripe task, which:

1. Automatically creates the "Premium tier" product.
2. Automatically creates the default billing portal configuration used in the /dashboard/manage-subscription route.
3. Prints the "Premium tier" product price ID to be copied into constants.ts.

This dramatically reduces the time and steps needed and the chance of errors in setting up Stripe.

Contributors, please feel free to test and provide feedback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,43 +73,11 @@ This will automatically configure the database tables and their settings for us.
 
 ### Payments
 
-1. Create the “Premium tier” product in Stripe via the Stripe CLI:
-
-```
-stripe products create \
-  --name="Premium tier" \
-  --default-price-data.unit-amount=500 \
-  --default-price-data.currency=usd \
-  --default-price-data.recurring.interval=month \
-  --description="Unlimited todos"
-```
-
-2. The resulting [product object](https://stripe.com/docs/api/products) will be
-   printed in the terminal. Copy the `default_price` value as
-   `STRIPE_PREMIUM_PLAN_PRICE_ID` in [`constants.ts`](constants.ts).
-
-3. Next, head over to
-   [your Stripe dashboard settings](https://dashboard.stripe.com/test/apikeys)
-   to copy the `STRIPE_SECRET_KEY` into your `.env` file. We recommend using the
-   test key for your development environment.
-
-4. Configure Stripe's customer portal settings:
-
-5. Ensure your users can manage their subscription via Stripe's Customer portal
-   by going to
-   [your Customer Portal settings](https://dashboard.stripe.com/test/settings/billing/portal)
-   and toggling on:
-
-- Payment methods > Allow customers to view and update payment methods.
-- Cancellations > Cancel subscriptions
-- Subscriptions > Customers can switch plans (and select your relevant
-  subscription tiers)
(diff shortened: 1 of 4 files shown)


lino-levan (commented):
This looks fine as a v1, but I'm wondering if providing some flags would make sense? Not sure if it is in scope for this "getting started" script.

iuioiua:
> This looks fine as a v1, but I'm wondering if providing some flags would make sense?

What do you suggest?

> Not sure if it is in scope for this "getting started" script.

It was deliberately written to replace instructions with automatic setup. Why would that be?

init/stripe.ts
```diff
@@ -0,0 +1,70 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { Stripe } from "stripe";
+import { SITE_DESCRIPTION } from "@/constants.ts";
+import "std/dotenv/load.ts";
+import { stripe } from "@/utils/stripe.ts";
+
+async function createPremiumTierProduct(stripe: Stripe) {
+  return await stripe.products.create({
+    name: "Premium tier",
```
lino-levan, line 9: I'm wondering if the stuff in this object should be configurable? Does that make any sense?

lino-levan:
That seems good enough for now. LGTM

init/stripe.ts
iuioiua: Good suggestion. Though, we'll leave this as-is for now in the spirit of prioritising ease of setup over customizability.

README.md
niklasmtj, line 1: What if we keep the manual instructions but put the automated script first in the chapter. So people who would like to do those steps manually have a guidance.

niklasmtj (approved):
Looks good! I think this will prevent a lot of headache for people starting out with Stripe

README.md
```diff
@@ -71,68 +71,39 @@ This will automatically configure the database tables and their settings for us.
...
+   recommend using the test key for your development environment.
+2. Run `deno task init:stripe` and follow the instructions. This automatically
+   creates your "Premium tier" product and configures the Stripe customer
+   portal.
 
-3. Next, head over to
-   [your Stripe dashboard settings](https://dashboard.stripe.com/test/apikeys)
-   to copy the `STRIPE_SECRET_KEY` into your `.env` file. We recommend using the
-   test key for your development environment.
+> Note: go to [init/stripe.ts] if you'd like to learn more about how the
```
iuioiua, line 82: @niklasmtj, regarding https://github.com/denoland/saaskit/pull/93#discussion_r1168232989, I'd rather have one or the other. I've added this comment to point people in the right direction if they want to learn more. At some point, I'll include technical details as JSDoc comments in the implementation to give an even more precise idea of what's happening. Do you think this addresses your concern?

niklasmtj: Sounds good to me 👍🏼

iuioiua:
I just tested this locally, and it works! 🥳

PR #99, opened 2023-04-17
feat: `deno task ok`

iuioiua:
I decided to try an idea I just came up with. Reviews and opinions from all are welcome.

The benefit of this task is that it provides a holistic check that a contributor can use before submitting a PR. The trade-off is that the contributor loses some visibility into which sub-task causes failure when a failure occurs.

This change also includes some other documentation improvements.


lambtron:
if this is an established pattern in the OSS community then i'm for it!! anything that makes getting started/contributing as easy as possible is +1

kt3k:
I see the benefit of this command. I'm in favor.

> if this is an established pattern in the OSS community

I sometimes see this pattern especially in small to middle sized project (ex. chalk https://github.com/chalk/chalk/blob/a370f468a43999e4397094ff5c3d17aadcc4860e/package.json#L23 ).

PR #105, opened 2023-04-18
feat: reset password flow

iuioiua:
This implementation is not yet working correctly. When I follow the process to reset my password, the process _appears_ to work, but when I log out and log back in, my password hasn't changed.

Stylistically, this code can be cleaned up a little, but this is an initial version.

@thorwebdev, are you able to take a look? Do you see anything that needs to be corrected with the implementation?


iuioiua:
I'll pause this for now in favour of more important Deno Hunt functionality.

Issue #103, opened 2023-04-18
proposal: Deno Hunt


iuioiua:
See the [Discord discussion](https://discord.com/channels/684898665143206084/1085986084653109438/1096255046561300482).

iuioiua:
The first version of Deno Hunt was done in #106. Further progress will be tracked in #60.

PR #106, opened 2023-04-20
feat: initial version of Deno Hunt and restructure

iuioiua:
The Deno SaaSKit demo is pivoting to becoming Deno Hunt instead of a to-do list app (see announcement [here](https://github.com/denoland/saaskit/releases/tag/v0.1.0)).

In the spirit of learning early, this PR sets the foundations of a Deno Hunt implementation while re-working some previously suboptimal aspects. The re-working focuses on reducing the number of moving parts required for workflows and bringing closer functionalities that work together. Please keep in mind that any change made here can be disputed, and questions are welcome.

Here's a summary of what's changed:
- `/api/subscription` route has changed to a more appropriately named `/api/stripe-webhooks`
- API endpoints have moved into their associated endpoints. E.g. POST method `/api/login` to `/login`, etc.
- To-do list logic has been removed and replaced with items, including a new table in the database.
- Some components have been moved into other component files. E.g. `<Header />` is only ever used in `<Layout />`, so it's been moved to the `Layout.tsx` file, etc.
- The dashboard has been removed in favour of the `/account` tree.
- A bit of restyling of inputs and buttons.

The following features are missing but can be added in proceeding PRs: comments, voting, author pages, item text, ranking algorithm, and item limit on the front page.

You can play around with the preview deployment by clicking the `View deployment` button below.

Towards #103


matthew-dean:
Just a question -- Deno SaasKit is self-explanatory and intuitive in what it is and used for. Deno Hunt is... I don't know what? The announcement doesn't really make this clear. Is it no longer a SaasKit? Or is, but additionally something else?

iuioiua:
SaaSKit is the set of tools used to create SaaS-based websites. Deno Hunt is an example of what you can build with SaaSKit and will be something like [Product Hunt](https://www.producthunt.com/) (hence the name) and [Hacker News](https://news.ycombinator.com/). Probably more like hacker news.

kt3k (approved):
LGTM

lambtron (approved):

PR #117, opened 2023-04-24
fix: incorrect time ago calculation

huai-jie:


iuioiua (approved):
Niiice. Thank you, @huai-jie. LGTM!

PR #120, opened 2023-04-26
feat: cleanup Supabase local dev config

iuioiua:
@thorwebdev, do you have any ideas on how we can further improve `supabase/config.toml`?


thorwebdev:
Ah sorry for the delay. No, generally the default config.toml should be fine 👍

PR #123, opened 2023-04-27
feat: use Deno KV for database

iuioiua:
This change also:
- Simplifies Stripe interaction logic by creating a Stripe user at signup instead of JIT later.
- Removes the need for `SupabaseAdminClient` and the `SUPABASE_SERVICE_ROLE_KEY` environment variable.
- Reduces the need for documentation as everything is handled within the codebase.

@lambtron, it's worth noting somewhere that if someone wants to check out the Supabase implementation of this, to check out https://github.com/denoland/saaskit/tree/3b1b14a97eef8859596015b22557d575d3b63c09

I also request mentioning somewhere that manual database setup isn't required as the database is "built-up dynamically" via KV calls or something like that.


utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+export async function createItem(initItem: InitItemValue) {
+  const itemId = crypto.randomUUID();
+
+  const itemKey = ["items", crypto.randomUUID()];
+  const itemsByUserKey = ["items_by_user", initItem.userId, itemId];
+
+  const item: ItemValue = { ...initItem, score: 0, createdAt: new Date() };
+  const res = await kv.atomic()
+    .check({ key: itemKey, versionstamp })
+    .check({ key: itemsByUserKey, versionstamp })
```
lucacasonato, line 21: Where is `versionstamp` coming from here? Did you mean to use `versionstamp: null`?

utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+export interface InitItemValue {
+  userId: string;
+  title: string;
+  url: string;
+}
+
+export async function createItem(initItem: InitItemValue) {
+  const itemId = crypto.randomUUID();
+
+  const itemKey = ["items", crypto.randomUUID()];
```
lucacasonato, line 15:
```suggestion
  const itemKey = ["items", itemId];
```

utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+  const res = await kv.atomic()
+    .check({ key: itemKey, versionstamp })
+    .check({ key: itemsByUserKey, versionstamp })
+    .set(itemKey, item)
+    .set(itemsByUserKey, item)
+    .commit();
+
+  if (res === null) {
+    throw new TypeError("Item with ID or for user already exists");
+  }
```
lucacasonato, line 28: Instead of erroring, why not run this in a loop until you find a UUID that isn't in use?

utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+  const commentsByUserKey = [
+    "comments_by_users",
+    initComment.userId,
+    commentId,
+  ];
+  const commentsByItemKey = ["comments_by_item", initComment.itemId, commentId];
+
+  const comment: CommentValue = { ...initComment, createdAt: new Date() };
+  const res = await kv.atomic()
+    .check({ key: commentsByUserKey, versionstamp })
```
lucacasonato, line 66: where is versionstamp coming from?

utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+  id: string;
+  stripeCustomerId: string;
+}
+
+export async function createUser(initUser: InitUserValue) {
+  const key = ["users", initUser.id];
+  const user: UserValue = { ...initUser, isSubscribed: false };
+
+  const res = await kv.atomic()
+    .check({ key, versionstamp })
```
lucacasonato, line 105: ditto

utils/db.ts
```diff
@@ -0,0 +1,146 @@
...
+    throw new TypeError("User with ID already exists");
+  }
+}
+
+export async function getUserByStripeCustomerId(stripeCustomerId: string) {
+  const iter = kv.list<UserValue>({ prefix: ["users"] });
+  for await (const res of iter) {
+    if (res.value.stripeCustomerId === stripeCustomerId) return res;
+  }
+}
```
lucacasonato, line 119: index users by customer id instead.

utils/db.ts
iuioiua: Yes, I'm using `versionstamp = null` for the time being for all instances of `versionstamp` within this file. Is it fine if I instead use some arbitrary value like `00000000000000010000` or similar?

utils/db.ts
iuioiua: Good catch. Done.

utils/db.ts
iuioiua: This was my quick and dirty workaround while thinking of alternatives. Either way, done now.

utils/db.ts
iuioiua: Done. Though, why inline over global? I assume it's because there's nothing in common between the various `versionstamp`s throughout the file. Having a single global suggests that there is and can be misleading. Is that right?

README.md
iuioiua, line 1: Nice

Issue #91, opened 2023-04-16
todo: rework HTTP redirects


iuioiua:
Spec:
- [ ] A HTTP redirect helper function for relative paths like `redirect(path: string, status: 300 | 301...) => Response`
- [ ] The HTTP redirects across the codebase were done without much thought from me and likely need to be corrected.

PR #131, opened 2023-04-28
http redirects function

andrewdelanoy:
#91 
Created redirect helper function, updated redirects in existing code to use function.

Only changed one redirect to put the user on the sign up page rather than login, those could probably still be looked at.


utils/redirect.ts
iuioiua, line 1: Are you able to:
1. Rename this file to `http.ts`?
2. Use the function signature outlined in the original issue (#91)? `redirect(path: string, status: 300 | 301...) => Response`
3. `body` can be `null`

twind.config.ts
iuioiua, line 1: Are you able to revert these changes?

routes/login/index.tsx
```diff
@@ -88,7 +93,10 @@ export default function LoginPage(props: PageProps) {
             Login with GitHub
           </OAuthLoginButton>
           <div class="text-center text-gray-500 hover:text-black mt-8">
-            <a href="/signup">Don't have an account? Sign up</a>
+            <a href="/signup">
+              Don't have an account?{" "}
+              <span style="color:rgba(190,24,93)">Sign up</span>
+            </a>
```
iuioiua, line 99: Can you pleas revert this change?

routes/account/manage.ts
```diff
@@ -3,6 +3,7 @@ import type { Handlers } from "$fresh/server.ts";
 import { stripe } from "@/utils/payments.ts";
 import type { AccountState } from "./_middleware.ts";
 import { getUser } from "@/utils/db.ts";
+import { RedirectHelper } from "../../utils/redirect.ts";
```
iuioiua, line 6: Can you please import from "@/utils/redirect.ts"?

iuioiua (changes requested):
Thanks for helping out, @andrewdelanoy! Could you please address my requests? Also, could you please:
1. Clean up any code commented out from this PR.
2. Ensure the HTTP status codes are correct according to each scenario? This was my main concern and what I ignored when implementing redirects as a whole.
3. Revert any changes that don't pertain to the primary purpose of this PR - HTTP redirects.

Thanks!

andrewdelanoy:
I will address all the changes you mentioned!

iuioiua:
I've just fixed a couple of merge conflicts. Please let me know if there's anything else I can help with.

iuioiua:
Hi @andrewdelanoy, I'd like to implement this in the next few days. If you cannot do so, are you happy for me to take this one? 🙂

huai-jie:
> 2. Ensure the HTTP status codes are correct according to each scenario? This was my main concern and what I ignored when implementing redirects as a whole.

Referencing https://deno.land/std@0.182.0/http/http_status.ts?source#L252 could be helpful.

iuioiua:
This has been superseded by #163. I appreciate your help either way, @andrewdelanoy!

PR #109, opened 2023-04-21
add parts of production docs

lambtron:
new production docs. parts are still missing such as:
- where to grab supabase api url
- digital ocean
- aws lightsail

cc @cdoremus

PR #136, opened 2023-04-29
docs: added prod deploy db content to readme

cdoremus:
This edit seemed to have slipped through the cracks from our previous Production Deployment `README.md` update. I was able to recover it from a local version of #109 that I had.


cdoremus:
@iuioiua told me that Deno KV replaces Supabase now as the SaaSKit database, so I'll close this PR.

PR #137, opened 2023-04-29
feat: add comment count

huai-jie:
add comment count


iuioiua:
Thanks again, @huai-jie. Frankly, I deliberately left this data out of `<ItemSummary />` because I thought of going the Product Hunt route and not mentioning the comment count. We'd have the link go to the item page instead of its `url` property. WDYT?

huai-jie:
Oh I see. I think comment count is quite useful info for me to know about how much discussion has been done for this particular item/product. 

Mentioning comment count cant represent how good this product is (vote definitely does) but it is the easiest way to know how many people give their thought on this product.

### Propose
Maybe we could add a tag label for the item. Only the Item with the discussion label shows the comment count.

PS: Just close the pull request if this merge is not useful. Im totally fine with it. No worry ;)

iuioiua (changes requested):
Actually, let's go ahead with this. However, this can be simplified by adding a prop to `<ItemSummary />` and using `items.length`. There's also `pluralize()`, which can be used.

huai-jie:
> Actually, let's go ahead with this. However, this can be simplified by adding a prop to `<ItemSummary />` and using `items.length`. There's also `pluralize()`, which can be used.

Sorry, I fail to see there is easier way to implement this. Could you give me some more details about it? I would like to try it out.

Case: `ItemSummary` required comment data to show count

I direct get the comment data as well from the `getItems()` function as the homepage does not acquire any comment data previously. 

The comment data have been acquired as `commentsRes` on the item page previously & now it is also used as `comment_count` property with `commentsRes.length` and then props to `ItemSummary`

Add the optional comment_count type so that it could be direct used as `props.value.comment_count` on `ItemSummary` of both the homepage & item page

iuioiua:
Sorry, ignore what I said about `items.length`. Instead of forming a join and counting comments, perhaps we can do one of the following:
1. Have a `commentsCount` (note: in camelCase) property for the item, which increases each time a comment is created.
2. Have another index, say `comments_count_by_item`, which increases each time a comment is created via a mutation (example [here](https://deno.com/blog/kv#use-cases-and-examples)).

Either way, the value can increase each time a comment is created. I'd prefer option 1, as I think it's simplest. WDYT?

huai-jie:
> Sorry, ignore what I said about `items.length`. Instead of forming a join and counting comments, perhaps we can do one of the following:
> 
> 1. Have a `commentsCount` (note: in camelCase) property for the item, which increases each time a comment is created.
> 2. Have another index, say `comments_count_by_item`, which increases each time a comment is created via a mutation (example [here](https://deno.com/blog/kv#use-cases-and-examples)).
> 
> Either way, the value can increase each time a comment is created. I'd prefer option 1, as I think it's simplest. WDYT?

Both sound good to me. Okay, then I would go forward with option 1.

huai-jie:
I have a concern. 

Do I need to take care of previous data? The current code not handling any count of the previous comments. 

And it may also cause issues like https://github.com/denoland/saaskit/issues/143 due to incompatible data in KV.

iuioiua:
I was going to say this is fine, as we're currently in beta. However, when devs fork and use this codebase, having this manual increment property would be a pain point when making database modifications. Hm... Good thing you brought it up.

Sorry to have you change the approach for this so many times. Handling relational data in a KV datastore is not trivial.

My final proposal (hopefully) would be to have some `getItemCommentsCount()` and add a `commentCount` property to `<ItemSummary />`. WDYT?

@kt3k, do you agree?

huai-jie:
Does it okay if I direct include it in `getAllItems()`?

```
export async function getAllItems(options?: Deno.KvListOptions) {
  const iter = await kv.list<Item>({ prefix: ["items"] }, options);
  const items = [];
    items.push(res.value);
      let commentsCount = 0;
      for await (
        const _ of kv.list(
          { prefix: ["comments_by_item", res.key[1]] },
        )
      ) {
        commentsCount++;
      }
      res.value = { commentsCount, ...res.value };
      items.push(res);
    }
  }
  return items;
}
```

iuioiua:
I'd prefer not. It seems too complicated. How about a separate function that accepts item ID and returns comment count? Then, it can be used on individual item pages.

huai-jie:
> I'd prefer not. It seems too complicated. How about a separate function that accepts item ID and returns comment count? Then, it can be used on individual item pages.

Ya, I have separated the function that returns `commentsCount` to `getItemCommentsCount()`

I use join `{commentsCount: comments.length, ...item}` on individual item page though 

Since individual item page already acquires the comment data,

there is no need to read the `commentsCount` again from `getItemCommentsCount()`

components/ItemSummary.tsx
```diff
@@ -29,7 +29,9 @@ export default function ItemSummary(props: Item) {
         {pluralize(props.score, "point")} by {props.userId}{" "}
         {timeAgo(new Date(props.createdAt))} ago •{" "}
         <a href={`/item/${props.id}`}>
-          Comments
+          {props.commentsCount
+            ? pluralize(props.commentsCount, "Comment")
+            : "Comments"}
```
iuioiua, line 34:
```suggestion
          {pluralize(props.commentsCount, "comment")}
```
`pluralize()` takes care of `commentsCount = 0`, and best to use lowercase "c".

routes/item/[id].tsx
Comment (iuioiua) [inline routes/item/[id].tsx:1]: It'd be cleaner to create a new `commentsCount` property instead of modifying the `item` property.

utils/db.ts
```diff
@@ -37,7 +43,12 @@ export async function createItem(initItem: InitItem) {
...
   const items = [];
-  for await (const res of iter) items.push(res.value);
+  for await (const res of iter) {
+    const { id } = res.value;
+    const commentsCount = await getItemCommentsCount(id);
+    res.value = { commentsCount, ...res.value };
+    items.push(res.value);
+  }
   return items;
 }
```
iuioiua, line 53: I prefer to leave this function alone and instead map up comment counts within the `GET /` handler. You can see how I did this in #144.

iuioiua (changes requested):
Nice. We're heading in the right direction now!

components/ItemSummary.tsx
huai-jie: From what I saw from this code it would be resulting '0 comments' 

```
export function pluralize(unit: number, label: string) {
  return unit === 1 ? `${unit} ${label}` : `${unit} ${label}s`;
}
```
Could I modify the function with 
```
unit > 1  ? `${unit} ${label}s` : `${unit} ${label}`
```

iuioiua: What would `pluralize(0, "comment")` return?

iuioiua: > From what I saw from `pluralize()` code it would be resulting '0 comments'

Yes, that's correct and expected.

iuioiua: For a further context, see https://www.quora.com/Why-do-we-say-use-a-plural-noun-when-we-specify-quantity-as-zero

huai-jie: Wow I just know that 😂. 

Good to know though.

iuioiua (approved):
LGTM! Thank you, @huai-jie. Very much appreciated.

PR #138, opened 2023-04-30
revert: filter middleware routes

iuioiua:
This caused email/password users to loose their respective `user` objects. I'll re-visit this at some point.


iuioiua:
This appears to no longer be an issue, thanks to #139.

Issue #133, opened 2023-04-28
bug: OAuth logins lack the `user` object


iuioiua:
This is because the `user` object is created in the POST `/signup` route, which OAuth signup/sign-in never hits.

iuioiua:
This appears to no longer be an issue, thanks to https://github.com/denoland/saaskit/pull/139.

PR #139, opened 2023-05-01
fix/refactor: simplify db types and re-introduce `getOrCreateUser()`

iuioiua:
This change:
* Simplifies how database objects are get and set by including the `id` property in the respective object. This means we can return the object's value, including the `id`, ignoring the key.
* Re-introduces `getOrCreateUser()`, which ensures a user object always exists. This is particularly valuable for OAuth users.

Fixes:
* #133

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -1,127 +1,132 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import { stripe } from "./payments.ts";
+
 export const kv = await Deno.openKv();
 
-export interface InitItemValue {
+interface InitItem {
   userId: string;
   title: string;
   url: string;
 }
 
-export async function createItem(initItem: InitItemValue) {
-  let res = null;
-  while (res === null) {
-    const itemId = crypto.randomUUID();
+export interface Item extends InitItem {
+  id: string;
+  createdAt: Date;
+  score: number;
+}
 
-    const itemKey = ["items", itemId];
-    const itemsByUserKey = ["items_by_user", initItem.userId, itemId];
+export async function createItem(initItem: InitItem) {
+  let res = { ok: false };
+  while (!res.ok) {
+    const id = crypto.randomUUID();
+    const itemKey = ["items", id];
+    const itemsByUserKey = ["items_by_user", initItem.userId, id];
+    const item: Item = { ...initItem, id, score: 0, createdAt: new Date() };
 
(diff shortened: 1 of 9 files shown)


utils/db.ts
```diff
@@ -132,24 +137,24 @@ export async function setUserSubscription(
...
   const key = ["users", id];
-  const userRes = await kv.get<UserValue>(key);
+  const userRes = await kv.get<User>(key);
 
   if (userRes.value === null) throw new Error("User with ID does not exist");
 
   const res = await kv.atomic()
     .check(userRes)
-    .set(key, { ...userRes.value, isSubscribed } as UserValue)
+    .set(key, { ...userRes.value, isSubscribed } as User)
```
kt3k, line 146: This seems only updating the user at `users, id` key and the other user data at `users_by_stripe_customer` is not update. I don't think saving the same entity (user) in 2 different places are good idea (That will be bug prone, and the broken data will be hard to fix.)

utils/db.ts
```diff
@@ -1,127 +1,132 @@
...
 
-export async function getUserIdByStripeCustomerId(stripeCustomerId: string) {
-  const res = await kv.get<UserValue["id"]>([
-    "user_ids_by_stripe_customer",
+export async function getUserById(id: string) {
+  const res = await kv.get<User>(["users", id]);
+  return res.value;
+}
+
+export async function getUserByStripeCustomerId(stripeCustomerId: string) {
```
kt3k, line 127: Let's make this function call `kv.get` twice, such as:
- `kv.get("user_id_by_stripe_customer_id", stripeCustomerId)` -> userId
- `kv.get("users", userId)` -> User

iuioiua: Done.

kt3k (approved):
LGTM

kt3k:
@iuioiua 
Not directly the issue of this change, but `comments` and `items` are also saved in 2 keys and I think those will be hard to manage in the future (especially when we allow users to update them).

I think we should store only item ids in `item_ids_by_user_id, <user_id>, item_id` key and get them like:
- `kv.list({ prefix: ["item_ids_by_user_id", userId] })` -> list of itemIds
- `kv.getMany([["items", itemId0], ["items", itemId1], ...])` -> list of Items

iuioiua:
Yes, this is the challenge of working with relations within a key-value store. A few clever utilities should solve this. I'll see what I can come up with.

PR #144, opened 2023-05-02
feat: user display names

### LINKED ISSUE #116: proposal: user display names
This can be a unique username or a non-unique display name for each user. It's worth ignoring what Hacker News or Product Hunt does here and instead asking which is better.

We should use a new table if it's to be a unique username. If it's to be a non-unique display name, it might be better to come from the `user_metadata` object.

I'd like to hear some thoughts.

CC @lambtron and @thorwebdev

iuioiua:
This will only work for new users. A database reset will likely be needed.

Closes #116.


iuioiua:
PTAL, @kt3k.

routes/item/[id].tsx
```diff
@@ -33,8 +38,18 @@ export const handler: Handlers<ItemPageData, State> = {
     }
 
     const comments = await getCommentsByItem(id);
+    const commentsUsers = await getUsersByIds(
+      comments.map((comment) => comment.userId),
+    );
```
Comment (kt3k) [inline routes/item/[id].tsx:43]: Nice 👍

kt3k (approved):
LGTM

PR #154, opened 2023-05-03
feat: add upvote function

huai-jie:
The code is really messy ATM 

& 

I encountered some issues while attempting to do this

Issues:
- page required to refresh to get the latest change. (each time you upvote/remove upvote the item, the page is refreshed)
- upvote change is not synchronized.

Any suggestions on how to improve this are really welcome and appreciated.


iuioiua (changes requested):
Wow, I like this. I've been wondering how we'd do this, but this provides a clearer picture and is a great foundation to build upon. I feel like this PR might take a bit of back-and-forth before it's ready, so we'll have to exercise some patience 😅

Either way, the common theme in my suggestions is that we can break up some of the logic of this functionality into smaller, simpler pieces.

routes/api/vote.ts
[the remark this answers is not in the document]
huai-jie: Ya I passing it with URL parameters because it is more straightforward. 

Form fields are a better option though if data like item.id & vote.id is sensitive

huai-jie:
Touching grass 🌳

Would come back to continue this afterward 👾

routes/item/[id].tsx
```diff
@@ -76,6 +81,8 @@ export default function ItemPage(props: PageProps<ItemPageData>) {
           <ItemSummary
             item={props.data.item}
             commentsCount={props.data.comments.length}
+            votes={props.data.votes}
+            curUserId={props.data.session?.user.id}
```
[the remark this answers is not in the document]
Comment (huai-jie) [reply routes/item/[id].tsx]: `curUserid` is not required but `vote.id` is required for the delete action.

I would just pass `votes` property for now.

islands/VoteBtn.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const voteId = props.votes.find((item) => item.itemId === props.item.id)?.id;
+  return (
+    <button
+      class={`cursor-pointer mr-2 ${
+        upvoted ? "text-pink-700" : "text-gray-300"
+      }`}
+      type="submit"
+      onClick={() => {
+        const url = new URL("/api/vote", location.href);
+        url.searchParams.set("to", props.item.id);
```
iuioiua, line 18:
```suggestion
        const url = `/api/vote?item_id=${props.item.id}`;
```

islands/VoteBtn.tsx
```diff
@@ -0,0 +1,39 @@
...
+        let method = "POST";
+        if (upvoted) {
+          url.searchParams.set("vote", voteId as string);
```
iuioiua, line 21:
```suggestion
```

islands/VoteBtn.tsx
```diff
@@ -0,0 +1,39 @@
...
+  return (
+    <button
+      class={`cursor-pointer mr-2 ${
+        upvoted ? "text-pink-700" : "text-gray-300"
+      }`}
+      type="submit"
+      onClick={() => {
+        const url = new URL("/api/vote", location.href);
+        url.searchParams.set("to", props.item.id);
+        let method = "POST";
```
iuioiua, line 19:
```suggestion
        const method = upvoted ? "DELETE" : "POST";
```

islands/VoteBtn.tsx
```diff
@@ -0,0 +1,39 @@
...
+        fetch(url, {
+          method,
+          credentials: "same-origin",
+        }).then((res) => {
+          if (res.ok) {
+            location.assign(window.location.href);
+          } else {
+            location.assign("/login");
+          }
+        });
```
iuioiua, line 33: We don't need this entire `.then()` method as it is now. The endpoint itself will take care of any unauthenticated requests itself. Instead, the response received from `fetch()` should be used to set the value of the `upvoted` variable.

routes/api/vote.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    const params = new URL(req.url).searchParams;
+    const itemId = params.get("to");
+
+    if (itemId) {
+      await createVote({
+        userId: ctx.state.session.user.id,
+        itemId,
+      });
+    }
```
iuioiua, line 20: We should first guard against if `itemId` is `null` by returning a `new Response(null, { status: 400 })`.

routes/api/vote.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    return Response.json({ ok: true });
```
iuioiua, line 22:
```suggestion
    return new Response(null, { status: 201 });
```

routes/api/vote.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    if (itemId && voteId) {
+      await deleteVote({
+        userId: ctx.state.session.user.id,
+        itemId,
+        voteId,
+      });
+    }
+
+    return Response.json({ ok: true });
```
iuioiua, line 41:
```suggestion
    return new Response(null, { status: 204 });
```

routes/api/vote.ts
```diff
@@ -0,0 +1,43 @@
...
+    const itemId = params.get("to");
+    const voteId = params.get("vote");
+
+    if (itemId && voteId) {
+      await deleteVote({
+        userId: ctx.state.session.user.id,
+        itemId,
+        voteId,
+      });
+    }
```
iuioiua, line 39: Ditto with guarding against when `itemId` is `null`.

routes/index.tsx
```diff
@@ -9,24 +9,47 @@ import {
...
+      return -1;
+    }
+    if (x < y) {
+      return 1;
+    }
+    return 0;
+  };
+  return items.sort(fn);
+};
+
```
iuioiua, line 40: Sorry, I should've been clearer. Can you please move the `compareFn` out into its own function? For now, I suggest  `compareScore()` and removing the `property` parameter. We can expand functionality later if needed.

utils/db.ts
```diff
@@ -100,6 +100,73 @@ export async function getItemCommentsCount(itemId: string) {
...
+export interface Vote extends InitVote {
+  id: string;
+  createdAt: Date;
+}
+
+export async function createVote(initVote: InitVote) {
+  let res = { ok: false };
+  while (!res.ok) {
+    const id = crypto.randomUUID();
+    const votesByUserKey = ["vote_by_users", initVote.userId, id];
```
iuioiua, line 118:
```suggestion
    const votesByUserKey = ["votes_by_users", initVote.userId, id];
```
WDYT? Other referencing to this index will need also to be adjusted.

utils/db.ts
```diff
@@ -100,6 +100,73 @@ export async function getItemCommentsCount(itemId: string) {
   return count;
 }
 
+interface InitVote {
+  userId: string;
+  itemId: string;
+  voteId?: string;
```
iuioiua, line 106: Hm... I'm starting to think that the `voteId` property is not needed. If the entry exists for a given user or item, I think that's all that's needed. We could instead have the value of the entry be something arbitrary, simple but not misleading (not `null`). Perhaps, `undefined`? We'd also have to comment on the arbitrary nature of the value. WDYT?

islands/VoteBtn.tsx
```diff
@@ -0,0 +1,39 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { Item, Vote } from "@/utils/db.ts";
+
+export default function VoteBtn(
+  props: { item: Item; votes: Vote[] | [] },
+) {
+  const upvoted = !!props.votes.filter((item) => item.itemId === props.item.id)
+    .length;
+  const voteId = props.votes.find((item) => item.itemId === props.item.id)?.id;
```
iuioiua, line 9: To further clarify, I think this island should have an `isVoted` parameter that is determined in the handler. The property could then be fed to a [local Preact signal](https://preactjs.com/guide/v10/signals/#local-state-with-signals), which is then updated according to responses from `/api/vote`.

iuioiua (changes requested):
A step in the right direction!

iuioiua:
The CI error appears to be unrelated to this change.

Also, can you please change the name of `VoteBtn` to `VoteButton` to align it with other component names?

utils/db.ts
huai-jie: LGTM! 

Ya I would replace all `vote_by_users` with `votes_by_users`

utils/db.ts
huai-jie: Ya, `userId` and `itemId` are enough to indicate that a vote has been cast.


Hmm.. I am not sure I understand this. Could you explain it in more detail or show me some examples?
>  We could instead have the value of the entry be something arbitrary, simple but not misleading (not null). Perhaps, undefined?

iuioiua: The value would be arbitrary because it's never read. Whether the key exists is all we care about. Hence, the value can be `undefined`.

islands/VoteBtn.tsx
huai-jie: I use `useState` in this case

utils/db.ts
huai-jie: I am currently removing `voteId` since it appears to have no applicable use cases ATM.

routes/api/vote.ts
```diff
@@ -0,0 +1,52 @@
...
+  async DELETE(req, ctx) {
+    return await (responseFn(req, ctx));
+  },
+};
+
+const responseFn = async (
+  req: Request,
+  // deno-lint-ignore no-explicit-any
+  ctx: HandlerContext<PageProps<any>, State>,
+) => {
```
iuioiua, line 20: Could you please make this a normal function? For consistency-sake.

routes/api/vote.ts
```diff
@@ -0,0 +1,52 @@
...
+  if (!ctx.state.session) {
+    return new Response(null, { status: 400 });
+  }
+
+  const params = new URL(req.url).searchParams;
+  const itemId = params.get("item_id");
```
iuioiua, line 26:
```suggestion
  const itemId = new URL(req.url).searchParams.get("item_id");
```

routes/api/vote.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const handler: Handlers<PageProps, State> = {
+  async POST(req, ctx) {
+    return await (responseFn(req, ctx));
+  },
+
+  async DELETE(req, ctx) {
+    return await (responseFn(req, ctx));
+  },
+};
```
iuioiua, line 14:
```suggestion
export const handler: Handlers<PageProps, State> = {
 POST: responseFn,
 DELETE: responseFn,
};
```

routes/index.tsx
```diff
@@ -9,25 +9,43 @@ import {
...
+  const y = Number(b.score);
+  if (x > y) {
+    return -1;
+  }
+  if (x < y) {
+    return 1;
+  }
+  return 0;
+};
+
```
iuioiua, line 36: Ditto. Could you please make this a normal function?

components/ItemSummary.tsx
```diff
@@ -32,8 +34,10 @@ export default function ItemSummary(props: ItemSummaryProps) {
         </span>
       </div>
       <div class="text-gray-500">
-        {pluralize(props.item.score, "point")} by{" "}
-        {getUserDisplayName(props.user)}{" "}
+        <span id={`score-${props.item.id}`}>
+          {pluralize(props.item.score, "point")}
+        </span>{" "}
+        by{"  "}{getUserDisplayName(props.user)}{" "}
```
iuioiua, line 40:
```suggestion
        by{" "}{getUserDisplayName(props.user)}{" "}
```

components/ItemSummary.tsx
```diff
@@ -32,8 +34,10 @@ export default function ItemSummary(props: ItemSummaryProps) {
         </span>
       </div>
       <div class="text-gray-500">
-        {pluralize(props.item.score, "point")} by{" "}
-        {getUserDisplayName(props.user)}{" "}
+        <span id={`score-${props.item.id}`}>
+          {pluralize(props.item.score, "point")}
+        </span>{" "}
```
iuioiua, line 39: This `<span />` with `id` attribute can be removed. See below.

islands/VoteButton.tsx
```diff
@@ -0,0 +1,42 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { Item, Vote } from "@/utils/db.ts";
+import { useState } from "preact/hooks";
+
+export default function VoteButton(
+  props: { item: Item; votes: Vote[] | [] },
+) {
```
iuioiua, line 7: Could you please create a `VoteButtonProps` interface for these properties? I'd also like there to be an `isVoted` property, and it be calculated within the page handler.

islands/VoteButton.tsx
iuioiua, line 1: Once you've addressed this review, can I do some work on this island? I have some ideas of how it can be simplified.

routes/item/[id].tsx
```diff
@@ -18,15 +18,18 @@ import {
...
   type User,
+  type Vote,
 } from "@/utils/db.ts";
 
 interface ItemPageData extends State {
   user: User;
   item: Item;
   comments: Comment[];
   commentsUsers: User[];
+  votes: Vote[];
```
Comment (iuioiua) [inline routes/item/[id].tsx:32]: Ditto. I'd like this to instead be an `isVoted` property.

utils/db.ts
```diff
@@ -100,6 +100,71 @@ export async function getItemCommentsCount(itemId: string) {
...
+    const { value } = await kv.get(votesByUserKey);
+    if (value) {
+      await kv.delete(votesByUserKey);
+      const itemKey = ["items", initVote.itemId];
+      const { value: item } = await kv.get<Item>(itemKey);
+      if (item) {
+        item.score && item.score--;
+        res = await kv.atomic().set(itemKey, item).commit();
+      }
+      return { status: 200 };
```
iuioiua, line 149:
```suggestion
```
No return object is needed.

iuioiua (changes requested):

islands/VoteButton.tsx
huai-jie: Ya sure

components/ItemSummary.tsx
huai-jie: I would remove this with `changePoint()` in `VoteButton.tsx` 

The point would not update though after clicking the `VoteButton` after removing this.

ps: You can access this code in the file history if needed.

islands/VoteButton.tsx
huai-jie: I think all the reviews have been addressed.

Please feel free to begin your work 👍

iuioiua:
@huai-jie, I've made a few improvements. PTAL and test. WDYT?

huai-jie:
LGTM!

iuioiua:
Thank you for all your work, @huai-jie!

PR #166, opened 2023-05-08
refactor: simplify and move `<PostCard />`

iuioiua:
This change:
1. Moves `<PostCard />` into `@/routes/blog/[slug].tsx` as it's the only place that `<PostCard />` is used.
2. Simplifies the signature of `<PostCard />`
3. Removes the US date format setting.


iuioiua:
PTAL, @kt3k.

kt3k:
> Removes the US date format setting.

I'd prefer to keep `en-US` (or any `en`) locale because most contents and UIs of the site is in English. Showing only date in different locale style can be strange.

iuioiua:
> I'd prefer to keep `en-US` (or any `en`) locale because most contents and UIs of the site is in English. Showing only date in different locale style can be strange.

Fair point. Fixed now.

kt3k (approved):
LGTM

PR #169, opened 2023-05-08
[experiment] refactor: remove comment count

iuioiua:
I'm playing with this idea, as the comment count may not be worth the added complexity.


iuioiua:
Closing in favour of #171.

PR #171, opened 2023-05-09
refactor: remove comments count

iuioiua:
This change removes the comments part of `<ItemSummary />`. [Previously](https://github.com/denoland/saaskit/pull/137), this was argued to be a good indicator of an interesting item. However, with the complexity and computational cost of such a minor feature and the onset of voting functionality, it's more reasonable not to have the feature. A vote score is a sufficient indicator of an item's value.

However, an item's comment count can still be seen on the item page, which is calculated without an extra database call through `comments.length`. A few other adjustments had to be made to make `<ItemSummary />` usable:
1. The item title now links to the item page.
2. The item URL host now links to the item URL in a new tab, with a symbol included to indicate so.
3. Both new links get underlined upon hovering to clarify their intent.

CC @huai-jie

### DIFF
--- a/components/ItemSummary.tsx
+++ b/components/ItemSummary.tsx
@@ -1,7 +1,6 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import VoteButton from "@/islands/VoteButton.tsx";
 import { getUserDisplayName, Item, User } from "@/utils/db.ts";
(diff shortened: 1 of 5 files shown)


huai-jie:
I'm good with it 👍

kt3k (approved):
LGTM

Issue #172, opened 2023-05-09
Bug: Fail to upvote others' posts


huai-jie:
**Describe the bug**
Fail to upvote others' posts

**To Reproduce**
Upvote posts that were not created by the owner

![image](https://user-images.githubusercontent.com/58794979/237015252-b40de336-3672-4dc6-8804-4249843327c2.png)

**Additional context** 
Sorry for missing this. I should have reviewed the code more thoroughly.

This code is looking weird ...
```
export async function createVote(initVote: InitVote) {
  const itemByUserKey = ["items_by_user", initVote.userId, initVote.itemId];
  const itemKey = ["items", initVote.itemId];
  const voteByUserKey = ["votes_by_users", initVote.userId, initVote.itemId];
```
`itemByUserKey` definitely fail to get 

since the `userId` from `initVote` is always `id` of the login user 

instead of owner/user of that particular item 

(if login user === item owner then it would work but won't if not ...)

PR #174, opened 2023-05-09
fix: fail to upvote others posts

huai-jie:
close https://github.com/denoland/saaskit/issues/172

### DIFF
--- a/routes/account/upgrade.ts
+++ b/routes/account/upgrade.ts
@@ -28,6 +28,6 @@ export const handler: Handlers<null, AccountState> = {
       mode: "subscription",
     });
 
-    return redirect(url || '/');
+    return redirect(url || "/");
   },
 };
--- a/routes/account/upgrade.ts
+++ b/routes/account/upgrade.ts
@@ -28,6 +28,6 @@ export const handler: Handlers<null, AccountState> = {
       mode: "subscription",
     });
 
-    return redirect(url || "/");
+    return redirect(url);
   },
 };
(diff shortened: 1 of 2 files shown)


utils/db.ts
```diff
@@ -140,20 +147,29 @@ export async function createVote(initVote: InitVote) {
...
 
   let res = { ok: false };
   while (!res.ok) {
-    const itemByUserRes = await kv.get<Item>(itemByUserKey);
     const itemRes = await kv.get<Item>(itemKey);
     const voteByUserRes = await kv.get<Item>(voteByUserKey);
 
+    if (itemRes.value === null) throw new Error("Item does not exist");
+
+    if (itemRes.value === null) throw new Error("Item does not exist");
```
iuioiua, line 160:
```suggestion
```

iuioiua (approved):

routes/account/upgrade.ts
```diff
@@ -28,6 +28,6 @@ export const handler: Handlers<null, AccountState> = {
       mode: "subscription",
     });
 
-    return redirect(url);
+    return redirect(url || "/");
```
iuioiua, line 31: Can you please revert this change?

huai-jie: Sure

it triggers this error though when running `deno task ok` on local machine

```
error: TS2345 [ERROR]: Argument of type 'string | null' is not assignable to parameter of type 'string'.
  Type 'null' is not assignable to type 'string'.
    return redirect(url);
                    ~~~
    at file:///....../...../saaskit/routes/account/upgrade.ts:31:21
```

iuioiua (approved):
LGTM! Good catch! Thank you, @huai-jie.

routes/account/upgrade.ts
iuioiua: @huai-jie, if you're still able to reproduce this issue, could you please create a GitHub issue?

PR #192, opened 2023-05-13
feat: Deno KV OAuth

iuioiua:
This change introduces an auth solution that uses Deno KV. The change is fundamental to the site but has resulted in a smaller code footprint and auth being "part of the package" of the codebase, meaning less setup and few external dependencies.

The new `/utils/deno_kv_oauth.ts` file is a small library that offers Deno KV + OAuth interplay. Once at a satisfactory standard, this might be pulled out into its own official module. Initially, I used an object-oriented approach, where the object was attached to the context state, `ctx.state`, in the site middleware. However, I found that a functional approach was much more straightforward. The only property on the context state is now a simple `sessionId?: string`.

Other related changes were needed as a result of the migration:
- The `redirect_to` parameter for logins has been removed.
- DB tests focusing on the user and sessions have been added, as they've become more vital to the site's functioning. More DB tests are to be added in the following PRs.
- All email/password-authentication-related functionality has been removed.


iuioiua:
The bugs are fixed on my local machine now.

utils/db.ts
```diff
@@ -185,33 +185,42 @@ export async function getVotedItemIdsByUser(
...
     .check({ key: usersKey, versionstamp: null })
-    .check({ key: stripeCustomersKey, versionstamp: null })
+    .check({ key: usersByLoginKey, versionstamp: null })
+    .check({ key: usersBySessionKey, versionstamp: null })
+    .check({ key: usersByStripeCustomerKey, versionstamp: null })
     .set(usersKey, user)
-    .set(stripeCustomersKey, user.id)
+    .set(usersByLoginKey, user)
+    .set(usersBySessionKey, user)
+    .set(usersByStripeCustomerKey, user)
```
kt3k, line 223: I don't think it's good idea to save the same user in 4 different keys (The number could even grow when we add a new field to User model). Why not saving field-to-user_id mapping like the existing `user_ids_by_stripe_customer`?

iuioiua: We previously used a `user_ids_by_stripe_customer` index that acts as a "join" between two indexes. However, the amount of coordination might be greater with that method. Saving and updating the same data across multiple indexes might be the way for KV databases. Other Deno KV implementations also follow this method (not that that's a real reason to follow).

kt3k: > However, the amount of coordination might be greater with that method.

Could you elaborate on this? Saving same data models in many keys looks more complex and causing more maintenance cost to me. Also it looks more error prone if someone forgot to update part of it in some update function.

> Other Deno KV implementations also follow this method

Can you give examples of this?

utils/deno_kv_oauth.ts
```diff
@@ -0,0 +1,110 @@
...
+import { assert } from "https://deno.land/std@0.178.0/_util/asserts.ts";
+
+const OAUTH_SESSION_COOKIE_NAME = "oauth-session";
+const OAUTH_SESSION_KV_PREFIX = "oauth-sessions";
+const SITE_SESSION_COOKIE_NAME = "session";
+
+interface OAuthSession {
+  state: string;
+  codeVerifier: string;
+}
```
kt3k, line 13: I think we also need `expireAt: number` field to expire it for inactivity.

kt3k:
I think we need to store and keep updating `expireAt` number for each session, and check if it's expired or not for each session request

utils/db.ts
iuioiua: > Could you elaborate on this? Saving same data models in many keys looks more complex and causing more maintenance cost to me. Also it looks more error prone if someone forgot to update part of it in some update function.

Say you have two indexes: `items` and `users`. If you need to get all items by a user, you'd probably have to have two intermediate indexes: `item_ids_to_user_ids` and `user_ids_to_item_ids`. This would mean you need to involve 2 indexes when getting all items by a user if you have the user ID. So, you have to find all items with that user ID using `kv.list()`, then get each item, likely using `kv.getMany()`, which is limited to 10 entries.

Compare this to the method used here. Using the user ID, you call `kv.list(["items_by_users", userId])` and you have all your items. To be clear, I'm not 100% sold on this method either, but so far, it seems most likely to be the best way to do it. As we discussed, it might be worth investing some time into a module that manages relational data within Deno KV better.

> Can you give examples of this?

- https://deno.com/manual@v1.33.3/runtime/kv/secondary_indexes
- https://github.com/denoland/tic-tac-toe
- https://github.com/hashrock/kv-sketchbook/blob/main/utils/db.ts

kt3k: hmm ok. Let's try multiple save strategy.

utils/db.ts
```diff
@@ -185,33 +185,42 @@ export async function getVotedItemIdsByUser(
...
   return voteItemIds;
 }
 
 interface InitUser {
   id: string;
+  login: string;
+  avatarUrl: string;
   stripeCustomerId: string;
-  displayName: string;
+  sessionId: string;
```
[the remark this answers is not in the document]
iuioiua: Hm... These are good points. I quickly played around with not having `sessionId` be a property of `User` and realised it mightn't be so easy. I want to tackle this problem in a follow-up PR focusing on this issue. I'd rather land this sub-optimal solution sooner than a perfect solution later.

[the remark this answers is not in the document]
kt3k: Ok. Makes sense

kt3k (approved):
LGTM

iuioiua:
I'll wait for 1 day until merging to allow for other reviews.

utils/db.ts
```diff
@@ -226,69 +235,175 @@ export async function getUserById(id: string) {
   return res.value;
 }
 
+export async function getUserByLogin(login: string) {
+  const res = await kv.get<User>(["users_by_login", login]);
+  return res.value;
+}
+
+export async function getUserBySessionId(sessionId: string) {
```
losfair, line 243: For sessions it would be good to attempt an eventually-consistent read first and fall back to strongly-consistent read if the first attempt returned not-found:

```ts
let res = await kv.get<User>(["users_by_session", sessionId], { consistency: "eventual" });
if(!res.value) {
  res = await kv.get<User>(["users_by_session", sessionId]);
}
return res.value;
```

utils/db.ts
```diff
@@ -226,69 +235,175 @@ export async function getUserById(id: string) {
...
+    user.stripeCustomerId,
+  ];
 
-  if (userRes.value === null) throw new Error("User with ID does not exist");
+  const [
+    userRes,
+    userByLoginRes,
+    userBySessionRes,
+    userByStripeCustomerRes,
+  ] = await Promise.all([
```
losfair, line 285: `kv.getMany`?

utils/db.ts
```diff
@@ -226,69 +235,175 @@ export async function getUserById(id: string) {
...
+  ];
 
-export async function getOrCreateUser(id: string, email: string) {
-  const user = await getUserById(id);
-  if (user) return user;
+  const [
+    userRes,
+    userByLoginRes,
+    userByStripeCustomerRes,
+  ] = await Promise.all([
```
losfair, line 334: `kv.getMany`

losfair (commented):
A few comments on more optimal usage of KV API

losfair (approved):
LGTM

lambtron (approved):

niklasmtj (approved):
Really good!

PR #193, opened 2023-05-14
refactor(typing): Add missing types in auth.ts

Sorikairox:
Add missing typing on`getRequestHeader`,  `getCookie` and `setCookie` function in auth.ts


iuioiua:
Hi @Sorikairox, I suggest pausing adding any more to this PR. Auth will soon use Deno KV instead of Supabase Auth. My initial thoughts are I wouldn't want you to put any further work into a feature that will soon be dropped.

Nevertheless, I like the direction of some of these changes! I'll do a proper review soon.

iuioiua:
Also, please don't forget to run `deno task ok` before committing 😄

Sorikairox:
@iuioiua I fixed it so `deno task ok` passes now ! But of course I will do it before creating a PR next time !

e2e/create-new-post.test.ts
iuioiua, line 1: I love the idea behind these tests. Few things regarding structure:
1. Could you please change the `e2e` folder name to `tests`? We'll probably move all tests to this folder in the future.
2. Could you please use `_` instead of `-` as the space separator for file names as per [the Deno style guide](https://deno.com/manual@v1.33.3/references/contributing/style_guide#use-underscores-not-dashes-in-filenames), and suffix any test files with `_test` (see `utils` folder for examples).
3. I'm happy to combine both test files into a single `tests/e2e_test.ts` file. WDYT?

iuioiua (changes requested):

iuioiua:
Deno KV OAuth is now implemented. I've changed the branch to be merged to `supbase-auth`, where the Supabase Auth implementation now lives.

Issue #199, opened 2023-05-15
bug: vote count becomes negative when you quickly press the vote button many times


brunocorrea23:
**Describe the bug** A clear and concise description of what the bug is.
Vote count becomes negative.
If you refresh the page, vote count goes back to how it was initially.

**To Reproduce** Steps to reproduce the behavior:

1. On the homepage, quickly press the upvote button many times
2. If you do this long enough, the vote count becomes negative
3. If you refresh the page, vote count goes back to how it was initially.

**Expected behavior** A clear and concise description of what you expected to
happen.
Vote count should never be negative.
Plus, a user should never be able to reduce the vote count, only to remove his/her own vote.

**Screenshots** If applicable, add screenshots to help explain your problem.
![Screenshot 2023-05-15 at 10 03 25](https://github.com/denoland/saaskit/assets/8139585/120313b7-b1aa-4682-b8c0-59cf3547314f)

iuioiua:
Good catch, @brunocorrea23! Upvoting/unvoting functionality deliberately adjusts the item's `score` property. The alternative would incur a heap of db calls, especially on the homepage, if calculating votes per item. This may change as Deno KV evolves in the future.

PR #200, opened 2023-05-15
fix: showing negative vote count

huai-jie:
close https://github.com/denoland/saaskit/issues/199


iuioiua (approved):
Elegant solution! Good stuff, @huai-jie!

PR #203, opened 2023-05-16
Append cookies in the response from supabase

### LINKED ISSUE #153: bug: logout is not working
Tested using OAuth. Clicking the logout button takes me to the homepage but doesn't log me out.

zaynetro:
Supabase client sets two cookies during logout:

```
set-cookie supabase-auth-token=; SameSite=Lax; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT
set-cookie supabase-auth-token-code-verifier=; SameSite=Lax; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT
```

Previously, with `set` method the first cookie was not returned to the browser.

Fixes #153

### DIFF
--- a/routes/_middleware.ts
+++ b/routes/_middleware.ts
@@ -38,6 +38,6 @@ export async function handler(
    * Note: ensure that a `new Response()` with a `location` header is used when performing server-side redirects.
    * Using `Response.redirect()` will throw as its headers are immutable.
    */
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
Good catch! Simple solution. Thank you, @zaynetro.

Issue #206, opened 2023-05-18
bug: fail to load account page after login


huai-jie:
**Describe the bug** 
Fail to load account page after login

**To Reproduce** 

1. Login
2. Click on 'Account'

**Expected behavior** 
Able to see Account detail

**Screenshots**
![image](https://github.com/denoland/saaskit/assets/58794979/ec03b213-a6e2-41e6-b611-e1251d1c35f8)

**Additional context** 
Happen due to missing `ctx.state.user`

PR #207, opened 2023-05-18
fix: cannot read properties of undefined (reading 'isSubscribed') on account page

huai-jie:
close https://github.com/denoland/saaskit/issues/206


iuioiua (approved):
Good catch! Thank you, @huai-jie.

PR #214, opened 2023-05-18
fix: move comment box to top

lino-levan:
Currently the comment box is at the bottom of the page, which isn't very intuitive. This PR just moves it to the top.

image

### DIFF
--- a/routes/item/[id].tsx
+++ b/routes/item/[id].tsx
@@ -100,6 +100,15 @@ export default function ItemPage(props: PageProps<ItemPageData>) {
             isVoted={props.data.isVoted}
             user={props.data.user}
           />
+          <form method="post">
+            <textarea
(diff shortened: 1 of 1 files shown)


iuioiua:
Yep, agreed. I think this needs to be done in conjunction with ordering comments from newest (top) to oldest (bottom). Are you able to do that too?

lino-levan:
I was going to suggest keeping the order as most upvoted, but I realize comments don't have upvotes (yet?). Is that planning on being added at some point? For this PR I can change the order of comments.

iuioiua:
Yeah, comments don't have upvotes. TBH, I'm not sure we'll add it, but that could change.

lino-levan:
Fair enough, I'll change the ordering right now to unblock the PR.

lambtron (approved):

Issue #211, opened 2023-05-18
proposal: complete homepage functionality


iuioiua:
This includes:
- [x] Pagination using a `?page=` URL query with items per page length of 10.

brunocorrea23:
@iuioiua , what do you believe is left to be done here in this issue?

iuioiua:
Item numbering, which should be easy.

iuioiua:
One more. Right now, we can only see items up till 1 week ago. It'd be nice to have a simple selector to change this. I think 1 week ago, 1 month ago and all-time ago are decent options. It could be controlled via a simple `?time-ago=` URL query.

iuioiua:
This was completed in #257.

PR #221, opened 2023-05-21
feat: homepage pagination and local db seeding

zzeleznick:
## Description
This PR intends to support https://github.com/denoland/saaskit/issues/211 by adding simple pagination with a `page` URL parameters and a helper script to seed the db with some dummy data.

### Adds
- `tools/seed_submissions.ts` script to pull in dummy data from HN for testing
- `page` option to specify the cursor (noting that there isn't yet a feed design that takes into account the date and score of the submissions)
- More button to navigate to the next page

### Example

Screen Shot 2023-05-23 at 9 30 40 PM

### Notes
- Did not add numbering as this would require passing the start index along with the cursor which seems like a good indicator for a design discussion
- Thinking there should be an iterator or indexer by "hotness" / "recency" as part of future work


routes/index.tsx
```diff
@@ -33,23 +34,28 @@ export function compareScore(a: Item, b: Item) {
...
 export const handler: Handlers<HomePageData, State> = {
-  async GET(_req, ctx) {
+  async GET(req, ctx) {
     /** @todo Add pagination functionality */
-    const items = (await getAllItems({ limit: 10 })).sort(compareScore);
+    const { searchParams } = new URL(req.url);
+    const limit = Math.min(30, ~~(searchParams.get("limit") || 10));
+    const start = decodeURI(searchParams.get("page") || '');
+    const { items: itemsRaw, cursor } = await getAllItems({ limit, cursor: start });
+    const items = itemsRaw.sort(compareScore);
```
iuioiua, line 43:
```suggestion
    const start = new URL(req.url).searchParams.get("page") || undefined;
    const { items, cursor } = await getAllItems({ limit: 10, cursor: start });
    items.sort(compareScore);
```
- The `start` constant can be calculated in one readable line.
- A hardcoded limit of 10 is fine.
- `Array.sort()` mutates the original array. So no need to create another array.

routes/index.tsx
```diff
@@ -33,23 +34,28 @@ export function compareScore(a: Item, b: Item) {
...
     /** @todo Optimise */
     const areVoted = items.map((item) => votedItemIds.includes(item.id));
-    return ctx.render({ ...ctx.state, items, users, areVoted });
+    return ctx.render({ ...ctx.state, items, cursor, users, areVoted });
   },
 };
 
 export default function HomePage(props: PageProps<HomePageData>) {
+  const nextUrl = new URL(props.url);
+  nextUrl.searchParams.set('page', props.data?.cursor || '');
```
iuioiua, line 58:
```suggestion
  const nextPageUrl = new URL(props.url);
  nextPageUrl.searchParams.set('page', props.data.cursor || '');
```

tools/seed_submissions.ts
iuioiua, line 1: While nice, things like performance measurements and batch handling promise arrays aren't necessary in this script, as it's just for generating dummy data on the local machine. No performance measurements, and sequential processing is fine 🙂

Either way, could you please convert the arrow functions to normal functions to match the rest of the codebase?

iuioiua (changes requested):
Nice approach! Please be sure to run `deno task ok` before pushing.

utils/db.ts
```diff
@@ -409,6 +412,16 @@ export async function deleteUserBySession(sessionId: string) {
...
+  for (const batch of batchify(keys, 10)) {
+    users.push(...(await kv.getMany<User[]>(batch)).map((entry) => entry.value!))
+  }
+  return users
+}
+
+export function* batchify<T>(arr: T[], n = 5): Generator<T[], void> {
+  for (let i = 0; i < arr.length; i += n) {
+    yield arr.slice(i, i + n);
+  }
```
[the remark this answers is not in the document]
zzeleznick: Can do. Originally `batchify` lived in the `/tools/submissions.ts` but I ran into issues when trying to call `kv.getMany()` with more than 10 keys (e.g. 30 when I expanded the range of values the page `limit` parameter would accept). Since we have made limit a fixed size of 10 again I will revert this.

iuioiua:
Mostly LGTM! However, I'm now getting the following error after running `deno task db:seed`:
```
An error occurred during route handling or page rendering. TypeError: Cannot read properties of null (reading 'login')
    at Object.ItemSummary (file:///Users/asher/GitHub/saaskit/components/ItemSummary.tsx:42:23)
    at https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1830
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1857)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1074)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:2844)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:2844)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1933)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1074)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1933)
    at T (https://esm.sh/v119/preact-render-to-string@5.2.6/X-ZS8q/deno/preact-render-to-string.mjs:12:1933)
```

zzeleznick:
@iuioiua – Yep. To minimize the diffs, I elected to make `User` optional as there is a flow in which a post can exist without a user in prod (e.g. user created, user posts, and then user is deleted) independently of the KV database population. LMK if this is reasonable.

iuioiua:
The user deletion flow is not yet defined or implemented. Even if it were, it'd likely delete all objects relating to the user, including posts.

tools/seed_submissions.ts
```diff
@@ -0,0 +1,83 @@
...
+async function seedSubmissions(stories: Story[]) {
+  const items = stories.map(({ by: userId, title, url }) => {
+    return { userId, title, url };
+  }).filter(({ url }) => {
+    try {
+      return Boolean(new URL(url).host);
+    } catch {
+      return;
+    }
+  });
```
iuioiua, line 66:
```suggestion
  }).filter(({ url }) => url);
```
Maybe just include posts with a `url` property.

zzeleznick:
@iuioiua – for context on my end, I saw a function and a corresponding test for user deletion -> https://github.com/denoland/saaskit/blob/main/utils/db_test.ts#L46

iuioiua:
Ah, yes. `deleteUser()` is currently a test function. I'll create a PR to move it to the test file. Thanks for pointing that out!

iuioiua (changes requested):

zzeleznick:
@iuioiua – Could you clarify what changes you need for this PR? 

For demo or local testing purposes, I feel like emulating real users via the `item.userId` as a fallback for the user ID works well.

iuioiua:
The changes to `<UserPostedAt />` and `<ItemSummary />` should be reverted. These components shouldn't be affected by this PR. Instead, the new seed submissions script should populate the `["users", id]` index. Again, the properties in the `User` object when creating the user are arbitrary. So these values don't have to be based on real data if it's too annoying to do so 😄

iuioiua:
Also, could you add a short explainer in the README and seed submissions script about what it does?

zzeleznick:
@iuioiua – PTAL when you have the chance
- [x] reverted changes to `<UserPostedAt />` and `<ItemSummary />`
- [x] created a dummy users with avatar URLs
- [x] updated the `Getting Started Locally` in `README.md`
- [x] updated the screenshot in the PR body

iuioiua (approved):
LGTM! Great work, @zzeleznick 💪🏾

PR #224, opened 2023-05-23
feat: record visit per day

huai-jie:


iuioiua (changes requested):
Mostly LGTM! I suggest just a few tweaks. Also, could you please create a simple `getVisitsPerDay(date: string): Promise<bigint>` (or similar) and add a test for both?

utils/db.ts
```diff
@@ -378,3 +378,23 @@ export async function getUsersByIds(ids: string[]) {
...
+}
+
+export async function getVisitsPerDay(date: Date): Promise<bigint> {
+  const visitsKey = [
+    "visits",
+    `${date.getUTCFullYear()}-${date.getUTCMonth()}-${date.getUTCDate()}`,
+  ];
+
+  return (await kv.get(visitsKey)).value as bigint;
+}
```
iuioiua, line 400:
```suggestion
export async function getVisitsPerDay(date: Date) {
  const res = await kv.get<bigint>([
    "visits",
    `${date.getUTCFullYear()}-${date.getUTCMonth()}-${date.getUTCDate()}`,
  ]);
  return res.value;
}
```
To make it consistent with other DB get functions.

utils/db_test.ts
```diff
@@ -85,4 +87,14 @@ Deno.test("[db] user", async () => {
...
+
+  const date = new Date("2023-01-01");
+  const visitsKey = [
+    "visits",
+    `${date.getUTCFullYear()}-${date.getUTCMonth()}-${date.getUTCDate()}`,
+  ];
+  await incrementVisitsPerDay(date);
+  assertEquals((await getVisitsPerDay(date)).valueOf(), 1n);
+  await kv.delete(visitsKey);
+  assertEquals(await getVisitsPerDay(date), null);
```
iuioiua, line 99: Can you please move this into its own separate test?

iuioiua (changes requested):

iuioiua (approved):
To clarify my previous point, I referred to a different `Deno.test()` within the same file. Either way, all good. I made the change 😄 

Thanks for your great work again, @huai-jie.

Issue #181, opened 2023-05-10
proposal: user pages


iuioiua:
It should:
- Be similar to the homepage in layout.
- Have a header including the user name, username and avatar image.
- Have a `/user/[username]` route.
- Be accessible by a new link on the `<ItemSummary />` component and comment component, where the username is.

iuioiua:
Implemented in #229.

PR #229, opened 2023-05-24
feat: user pages

brunocorrea23:
https://github.com/denoland/saaskit/issues/181

### DIFF
--- a/routes/user/[username].tsx
+++ b/routes/user/[username].tsx
@@ -1,0 +1,89 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { Handlers, PageProps } from "$fresh/server.ts";
+import Head from "@/components/Head.tsx";
+import Layout from "@/components/Layout.tsx";
+import { BUTTON_STYLES, NOTICE_STYLES } from "@/utils/constants.ts";
+import { ComponentChild } from "preact";
+import { getUserByLogin, type User } from "@/utils/db.ts";
+import type { State } from "@/routes/_middleware.ts";
+
+export interface UserData extends State {
+  user: User;
+}
+
+export const handler: Handlers<UserData, State> = {
+  async GET(_request, ctx) {
+    const { username } = ctx.params;
+
+    const user = await getUserByLogin(username);
+    if (user === null) {
+      return ctx.renderNotFound();
+    }
+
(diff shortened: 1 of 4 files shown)


routes/user/[username].tsx
```diff
@@ -0,0 +1,90 @@
...
+
+    const user = await getUserByLogin(username);
+    if (user === null) {
+      return ctx.renderNotFound();
+    }
+
+    return ctx.render({
+      ...ctx.state,
+      user: user!,
+    });
```
Comment (iuioiua) [inline routes/user/[username].tsx:26]:
```suggestion
    return ctx.render({ ...ctx.state, user });
```

routes/user/[username].tsx
```diff
@@ -0,0 +1,90 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { Handlers, PageProps } from "$fresh/server.ts";
+import Head from "@/components/Head.tsx";
+import Layout from "@/components/Layout.tsx";
+import { BUTTON_STYLES, NOTICE_STYLES } from "@/utils/constants.ts";
```
Comment (iuioiua) [inline routes/user/[username].tsx:5]:
```suggestion
```

routes/user/[username].tsx
```diff
@@ -0,0 +1,90 @@
...
+        )}
+      </div>
+    </li>
+  );
+}
+
+export default function UserPage(props: PageProps<UserData>) {
+  return (
+    <>
+      <Head title="User" href={props.url.href} />
```
Comment (iuioiua) [inline routes/user/[username].tsx:69]:
```suggestion
      <Head title={props.data.user.login} href={props.url.href} />
```

iuioiua (changes requested):
You're heading in the right direction! However, the page I had in mind had an avatar image and username at the top, aligned vertically or horizontally (whichever looks and feels nicer). The rest of the page would be a feed of the items by that user, using a new `getItemsByUser()` in `/utils/db.ts` and using `<ItemSummary />`.

Please let me know if you need any help with anything.

brunocorrea23:
nvm, already included the new getItemsByUser() in /utils/db.ts and also added the items to the user page

iuioiua (approved):
This is 🔥

We can de-duplicate some of the shared code between the user page and the home page in a follow-up PR. Thank you, @brunocorrea23.

Issue #202, opened 2023-05-15
proposal: fake business charts


lambtron:
the /dashboard page should contain fake charts:

![image](https://github.com/denoland/saaskit/assets/1080190/ff9f8d6a-cbc6-4890-88cf-83a33e8c2af2)
see https://next-fire-demo.makerkit.dev/dashboard

iuioiua:
I think these charts should exist on a `/stats` page. A few ideas of the charts we could have include:
- New items per day
- New votes per day
- New users per day
- Maybe even a visits per day?

iuioiua:
Visits per day were implemented in #224. The next step would be to create a `/stats` page with a simple but sexy full-page graph using this new data. I think we should use https://github.com/denoland/fresh_charts to do this. If anyone wants to tackle this, please let us know here.

brunocorrea23:
@iuioiua , should this be closed or are you thinking of something else?

iuioiua:
I'll close this now as it's implemented. We can organise polishing it off in #251. Thanks!

PR #235, opened 2023-05-27
feat: stats/charts page

brunocorrea23:
#202


iuioiua (approved):
Great work! I've made a few tweaks, which you can check out. Thank you, @brunocorrea23.

Issue #240, opened 2023-05-28
bug: month value of `visits` is saved incorrectly


huai-jie:
**Describe the bug** 
The `month` value is saved incorrectly in the database due to the zero-based indexing used by the `getUTCMonth()` method

**Screenshots** 
![image](https://github.com/denoland/saaskit/assets/58794979/0abe9386-60ee-406e-995e-b6178f8cb0f1)

PR #241, opened 2023-05-28
fix: incorrect month of visits

huai-jie:
close https://github.com/denoland/saaskit/issues/240

### DIFF
--- a/islands/VoteButton.tsx
+++ b/islands/VoteButton.tsx
@@ -24,7 +24,7 @@ export default function VoteButton(props: VoteButtonProps) {
       }
       isVoted.value = !isVoted.value;
       method === "POST" ? score.value++ : score.value--;
-      if (score.value < props.item.score) score.value = props.item.score;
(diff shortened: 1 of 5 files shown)


iuioiua (changes requested):
Hey @huai-jie, the issue is that it needs to be presented in a way that makes sense to humans. What if we format the display date using `Date.toLocaleDateString()` instead? Like "Dec 12, 2023" and "Jul 20, 2022". WDYT?

huai-jie:
I like this idea!

But I would still change the way to save the visit `date`.

Because `getUTCmonth()` utilize zero-based indexing, it would consistently cause a discrepancy with the actual month value.

iuioiua:
It looks like we'll need to clear the previous stats data. Is that correct?

huai-jie:
> It looks like we'll need to clear the previous stats data. Is that correct?

Ya ...

iuioiua:
I'll merge once `incrementVisitsPerDay()` is re-enabled.

iuioiua (approved):
LGTM! Thanks, @huai-jie.

PR #236, opened 2023-05-27
fix: pagination order

brunocorrea23:
#231

PR #242, opened 2023-05-29
feat: page-number-based pagination

### LINKED ISSUE #231: bug: pagination seems to break the order of Items
**Describe the bug**
Now the last item on the first page has zero upvotes.
The first item on the second page has two upvotes.
I guess this is happening because the ordering is being applied within the context of a page (ie: after the pagination split the items into pages). It should probably happen before.

**Expected behavior**
If a item on a second page has more upvotes than an item in the first page, it should be moved to the first page.

### LINKED ISSUE #233: proposal: pagination with numbers
Today a cursor is being passed on the URL params. Would it be better to have a number for the pagination?

A suggestion would be to pass the page number ('n') as an optional parameter to the getAllItems() function on the db.ts file. Then iterate on the list (with the the {limit: 10} option, for example) until you reach the n-th set of items.

This way index.tsx file wouldn't have to know anything about a cursor.

iuioiua:
This PR:
1. Introduces page-number-based pagination to the homepage.
2. Refactors much of the database-related code to be more straightforward.
3. Changes the homepage to show only the items posted in the last week.
4. Supercedes #236
5. Moves functionality closer to #211
6. Closes #233
7. Closes #231

Note: once merged, these same changes can be made to paginate item and user pages.

Please feel free to review, @brunocorrea23 and @huai-jie.

### DIFF
--- a/routes/api/stripe-webhooks.ts
+++ b/routes/api/stripe-webhooks.ts
@@ -2,7 +2,10 @@
 import type { Handlers } from "$fresh/server.ts";
 import { stripe } from "@/utils/payments.ts";
 import { Stripe } from "stripe";
-import { getUserByStripeCustomerId, setUserSubscription } from "@/utils/db.ts";
+import {
+  getUserByStripeCustomerId,
+  updateUserIsSubscribed,
+} from "@/utils/db.ts";
 
 const cryptoProvider = Stripe.createSubtleCryptoProvider();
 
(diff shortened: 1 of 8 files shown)


huai-jie:
Does this PR gonna tackle `New votes per day`  functionality that is mentioned in https://github.com/denoland/saaskit/issues/202?

iuioiua:
Not so much. That was tackled in #235.

brunocorrea23:
@iuioiua , thanks for the work here and for requesting the review.
Really liked the db refactor.

Couple of comments on the proposed solution.
- It seems that the items older than a week are never accessible to the users, right? Is this the intended behavior?
- 'Loading' all items for the last week provides no guarantee of the amount of items that will be 'loaded'. Although it is better than loading the whole db at once, I feel this kind of defeats the purpose of pagination.

Wdyt?

iuioiua:
> * It seems that the items older than a week are never accessible to the users, right? Is this the intended behavior?

Yes, as it stands now. Later, we can add functionality to make the window customisable.

> * 'Loading' all items for the last week provides no guarantee of the amount of items that will be 'loaded'. Although it is better than loading the whole db at once, I feel this kind of defeats the purpose of pagination.

True, the amount of loading is sub-optimal. This will be fixed if/when [`Deno.Kv.count()` is implemented](https://github.com/denoland/deno/issues/18965).

Does the PR look good to you?

brunocorrea23:
ok, thanks for the clarification!

LGTM

PR #244, opened 2023-05-30
feat: adding new stats

brunocorrea23:
- Refactored the code a little bit to make it very easy to add new stats
- Added missing stats from #202 (New items per day, New votes per day, New users per day)
- Changed the kv key for the 'visits per day' stats. Means we would lose the data we had before if there is no migration.
- UI for the /stats page needs some love
- Does not address #240 as it should be solved in #241

### DIFF
--- a/routes/index.tsx
+++ b/routes/index.tsx
@@ -10,7 +10,7 @@ import {
   getAllItemsInPastWeek,
   getAreVotedBySessionId,
   getManyUsers,
-  incrementVisitsPerDay,
+  incrementAnalyticsMetricPerDay,
   type Item,
   type User,
 } from "@/utils/db.ts";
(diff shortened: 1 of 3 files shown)


iuioiua:
Are you able to increase the spacing between graphs? They appear a little squished up.

iuioiua:
The new layout on desktop is great! Are you able to fix the layout on smaller screens?
localhost_8000_stats(iPhone SE)
localhost_8000_stats

brunocorrea23:
@iuioiua , I would appreciate some help here.
I noticed that the charts are not responsive, but I'm not sure how to fix this.
Can we tackle this on a next PR? I can create an issue. Also would like to add a tooltip, but also not sure how to do it.

iuioiua:
Ok, I'll merge this when I can. Let's tackle the responsiveness issue after 👍🏾

iuioiua (approved):
Sorry for the delay. After making a tweak, LGTM for now! Note: I do think that some of this logic is a little difficult to understand. We can address this in a follow-up PR, alongside increasing the polish of the entire codebase for v1.

PR #246, opened 2023-05-31
feat: adding item numbering on the homepage

### LINKED ISSUE #211: proposal: complete homepage functionality

brunocorrea23:
Should close #211

### DIFF
--- a/components/ItemSummary.tsx
+++ b/components/ItemSummary.tsx
@@ -4,6 +4,7 @@ import type { Item, User } from "@/utils/db.ts";
 import UserPostedAt from "./UserPostedAt.tsx";
 
 export interface ItemSummaryProps {
+  rank?: number;
   item: Item;
   user: User;
   isVoted: boolean;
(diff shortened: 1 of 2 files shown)


iuioiua:
Hm... Looking at it now, I wonder whether we should have this. Rank being so close to voting score clashes visually. What do you guys think, @brunocorrea23 and @kit?
image

huai-jie:
> and @kit?


Hmm... Looking at it, I wonder is asking me or not ? 🤔

JK

Yup I don't think item numbering is necessary for this case

brunocorrea23:
I think it's nice to have the rank of items. Reaching a higher rank gives an idea of an achievement. I think it's specially useful  when you are not in page number 1. 

I was inspired by Hacker News.
Their solution is visually better probably probably because vote count (or points) is smaller than the item rank.
Screenshot 2023-05-31 at 08 04 30

That being said, no strong conviction here. Either way looks good to me.

iuioiua:
Ok. I'll close this for now. This feature would be nice-to-have but not critical. Either way, much appreciated, @brunocorrea23!

PR #249, opened 2023-06-02
feat: redirect to referer page after login

huai-jie:
### i.e 
Clicking on the `comment` button on an item's page initiates the `login` process. It would be beneficial if, after logging in, users could be redirected back to the same item page.

### DIFF
--- a/routes/_middleware.ts
+++ b/routes/_middleware.ts
@@ -2,6 +2,7 @@
 import { MiddlewareHandlerContext } from "$fresh/server.ts";
 import { walk } from "std/fs/walk.ts";
 import { getSessionId } from "@/utils/deno_kv_oauth.ts";
+import { setCookie } from "std/http/cookie.ts";
 
 export interface State {
   sessionId?: string;
@@ -25,5 +26,15 @@ export async function handler(
 
   ctx.state.sessionId = getSessionId(req.headers);
 
-  return await ctx.next();
+  const res = await ctx.next();
+
+  if (ctx.destination === "route" && pathname === "/login") {
+    setCookie(res.headers, {
+      name: "redirectUrl",
+      value: req.headers.get("referer")!,
+      path: "/",
+    });
+  }
+
+  return res;
 }
--- a/routes/callback.ts
+++ b/routes/callback.ts
@@ -11,6 +11,7 @@ import { stripe } from "@/utils/payments.ts";
 import { State } from "./_middleware.ts";
 import { getAccessToken, setCallbackHeaders } from "@/utils/deno_kv_oauth.ts";
 import { oauth2Client } from "@/utils/oauth2_client.ts";
+import { deleteCookie, getCookies } from "std/http/cookie.ts";
 
 interface GitHubUser {
   id: number;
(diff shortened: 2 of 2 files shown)


iuioiua:
Interesting means of redirecting... I like it! You might be able to extract both code blocks into a single combined middleware function and apply this to the `/submit` route. Can you do that?

iuioiua:
Deno SaaSKit aims to be modular. Having the newly introduced functionality from this PR encapsulated in a single function would be ideal, perhaps using `pathnames: string[]` as a parameter. If you don't think this is feasible, fair enough. WDYT?

huai-jie:
Okay, I believe it's worth a try to implement this improvement. 

> Will try to do this whenever I am free.

iuioiua:
@huai-jie, can you do this for `/account` too? See #254 for context.

huai-jie:
> @huai-jie, can you do this for `/account` too? See #254 for context.

Sure 👌

utils/redirect.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function redirectToLogin(url: string) {
+  return redirect(`/signin?from=${url}`);
+}
+
+export function setRedirectUrlCookie(req: Request, res: Response) {
+  const from = new URL(req.url).searchParams.get("from");
+  setCookie(res.headers, {
+    name: REDIRECT_URL_COOKIE_NAME,
+    value: from ? from : req.headers.get("referer")!,
```
iuioiua, line 15:
```suggestion
    value: from ?? req.headers.get("referer")!,
```

utils/redirect.ts
```diff
@@ -0,0 +1,37 @@
...
+    path: "/",
+  });
+}
+
+export function setHeaderLocation(headers: Headers, url: string) {
+  headers.set("location", url);
+}
```
iuioiua, line 22: I don't think this function is needed.

utils/redirect.ts
```diff
@@ -0,0 +1,37 @@
...
+
+export function deleteRedirectUrlCookie(headers: Headers) {
+  deleteCookie(headers, REDIRECT_URL_COOKIE_NAME);
+}
+
+export function getRedirectUrlCookie(headers: Headers) {
+  const { redirectUrl } = getCookies(headers);
+  return redirectUrl;
```
iuioiua, line 30:
```suggestion
  return getCookies(headers)[redirectUrl];
```

utils/redirect.ts
```diff
@@ -0,0 +1,37 @@
...
+}
+
+export function setCallbackHeaders(req: Request, res: Response) {
+  const { redirectUrl } = getCookies(req.headers);
```
iuioiua, line 34:
```suggestion
  const redirectUrl = getRedirectUrlCookie(req.headers);
```

iuioiua (changes requested):
Could you please merge `@/utils/http.ts` into the new `@/utils/redirect.ts` along with tests?

utils/redirect.ts
```diff
@@ -0,0 +1,43 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { RedirectStatus, Status } from "std/http/http_status.ts";
+import { deleteCookie, getCookies, setCookie } from "std/http/cookie.ts";
+
+export const REDIRECT_URL_COOKIE_NAME = "redirectUrl";
```
iuioiua, line 5:
```suggestion
export const REDIRECT_URL_COOKIE_NAME = "redirect-url";
```
Last nit

Issue #227, opened 2023-05-24
proposal: trim Docker documentation


iuioiua:
I have no experience with Docker documentation, but I have some ideas of how it could be cleaned up:

- [x] Remove: https://github.com/denoland/saaskit/blob/f2364de163054ac99bb2bbf150c92159d9e73779/README.md?plain=1#L33
- [x] Create this file in the directory so the user doesn't have to: https://github.com/denoland/saaskit/blob/f2364de163054ac99bb2bbf150c92159d9e73779/README.md?plain=1#L188-L200
- [x] Ditto: https://github.com/denoland/saaskit/blob/f2364de163054ac99bb2bbf150c92159d9e73779/README.md?plain=1#L202-L210
- [x] Ditto: https://github.com/denoland/saaskit/blob/f2364de163054ac99bb2bbf150c92159d9e73779/README.md?plain=1#L212-L232
- [x] Add a step number to each instruction, like other sections in the README.

Please correct me if I'm wrong on any of these.

lambtron:
this makes sense to me. i can create a PR for it.

iuioiua:
Some of this was done in #238.

lambtron:
This PR addresses the remaining tasks in this issue: https://github.com/denoland/saaskit/pull/253

iuioiua:
This has been completed in #253.

PR #253, opened 2023-06-07
docs: add Docker files

lambtron:
Addresses https://github.com/denoland/saaskit/issues/227


Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM denoland/deno
+
+EXPOSE 8000
+
+WORKDIR /app
+
+ADD . /app
+
+RUN deno cache main.ts --import-map=import_map.json
```
iuioiua, line 9:
```suggestion
RUN deno cache main.ts
```
I'd assume the `import-map` flag isn't needed. Is that correct?

Dockerfile
iuioiua, line 1: This script doesn't run any script; it just caches dependencies. Shouldn't this be like the first script in the [Setting Up Docker](https://github.com/denoland/saaskit#setting-up-docker) section? I'd also trim the `CMD` to `deno task start`.

iuioiua (changes requested):
Can you remove the appropriate snippets these new files replace in the README?

iuioiua:
I made the changes based on my original suggestions. PTAL.

iuioiua (approved):
LGTM. @lambtron and @cdoremus, please review after the merge.

Issue #212, opened 2023-05-18
proposal: pagination on the item page for comments


iuioiua:
This should be done using a `?page=` URL query.

iuioiua:
For those wanting guidance, please check out https://github.com/denoland/saaskit/pull/221, which applied pagination to the homepage.

iuioiua:
This was completed in #257.

Issue #256, opened 2023-06-08
bug: item doesn't support more than 10 comments and throws a server 500 error


brunocorrea23:
Simply try to add more than 10 comments on any item.

For example, check: https://saaskit.deno.dev/item/7c47b301-17b5-4864-b7c5-e970d428877f

500 internal error: too many ranges (max 10)

<img width="431" alt="Screenshot 2023-06-08 at 10 52 00" src="https://github.com/denoland/saaskit/assets/8139585/ff1cd3cc-c56c-4f50-a918-18d0bf4ef746">

iuioiua:
This has been fixed in #257.

PR #257, opened 2023-06-08
feat: pagination on the item page for comments + time-ago filter

brunocorrea23:
#212 
#211


iuioiua:
Are you able to confirm whether this solves #256?

brunocorrea23:
In a way, yes. This PR allows for the user to add more than 10 comments in an item if you maintain PAGE_LENGTH=10.

But the core of the issue is still not solved.
It seems that the issue is connected to the Kv.getMany function on the getManyUsers function. It seems that the Kv.getMany function can't handle an array with more than 10 items as an input. (are you able to confirm if this is a limitation of the getMany function in kv?).

So if you slice the pages in groups of 10 elements (items, comments, etc) or less, you should be fine. But if you change PAGE_LENGTH to 20, for example, it throws an error. This is true not only for the items page, but also for the home page.

iuioiua:
Yes, `PAGE_LENGTH` should not increase beyond 10 because of this current limitation in Deno KV.

routes/index.tsx
```diff
@@ -54,25 +62,12 @@ export const handler: Handlers<HomePageData, State> = {
...
-          max={props.lastPage}
-          value={props.currentPage}
-          // @ts-ignore: this is valid HTML
-          onchange="this.form.submit()"
-        />
-        <label for="current_page" class="whitespace-nowrap align-middle">
-          of {props.lastPage}
-        </label>
-      </form>
+    <div class="flex justify-center">
```
iuioiua, line 67: Please add a note above saying these links do not preserve current URL queries. E.g. if `?page=2`, that'll be removed once one of these links is clicked. This behaviour is fine in this particular context but might need to change in the future as more URL queries control the item list view.

routes/item/[id].tsx
```diff
@@ -12,53 +13,66 @@ import {
...
+  lastPage: number;
+}
+
+function calcPageNum(url: URL) {
+  return parseInt(url.searchParams.get("page") || "1");
+}
+
+function calcLastPage(total = 0, pageLength = PAGE_LENGTH): number {
+  return Math.ceil(total / pageLength);
 }
```
Comment (iuioiua) [inline routes/item/[id].tsx:46]: Can we centralise and de-duplicate any pagination-related code in the utils folder with accompanying tests?

routes/item/[id].tsx
```diff
@@ -12,53 +13,66 @@ import {
...
   getUserBySessionId,
-  getVotedItemsByUser,
   type Item,
   type User,
 } from "@/utils/db.ts";
 import { redirect } from "@/utils/http.ts";
 import UserPostedAt from "@/components/UserPostedAt.tsx";
 import { pluralize } from "@/utils/display.ts";
 
+const PAGE_LENGTH = 10;
```
Comment (iuioiua) [inline routes/item/[id].tsx:29]: Please add a short explainer regarding why this should not exceed 10. Perhaps, there's a GitHub issue in the main https://github.com/denoland/deno repo that could be referenced also.

iuioiua (changes requested):
Sorry for the delay in the review. I've given it a quick read and left some suggestions.

iuioiua:
Hi @brunocorrea23, just checking in. Are you still happy to pursue this PR?

brunocorrea23:
Hey! Yes. Will tackle this today. Couldn't take a look over the week, sorry.

iuioiua (approved):
Pretty much LGTM! I think it might be best positioned above the item list, though. WDYT?

iuioiua (approved):
Awesome! LGTM! Thank you, @brunocorrea23.

Issue #251, opened 2023-06-05
todo: make charts responsive / mobile friendly


brunocorrea23:
![image](https://github.com/denoland/saaskit/assets/8139585/d9e51bda-72a6-445b-9145-700a6602e970)

PR #259, opened 2023-06-13
fix: charts not responsive

huai-jie:
close https://github.com/denoland/saaskit/issues/251

Before
image

After
image


static/app.css
iuioiua, line 1: Could this instead be a `<style>` in the `<Head />` for just the stats page?

iuioiua (changes requested):
Can we insert the new class in a `<style>` in the `<Head />` for the stats page only? We should also note somewhere that this is a workaround for [Fresh Charts not being responsive](https://github.com/denoland/fresh_charts/issues/11).

iuioiua:
Sorry, @huai-jie. To clarify, I'm happy to merge this PR, as long as we note that it's a workaround.

huai-jie:
> Can we instead insert the new class in a `<style>` in the `<Head />` for the stats page? We should also note somewhere that this is a workaround for [Fresh Charts not being responsive](https://github.com/denoland/fresh_charts/issues/11). I've made a suggestion for that issue.

done it

iuioiua (approved):
LGTM! Thank you, @huai-jie!

PR #263, opened 2023-06-17
fix: bigint serialization error

### LINKED ISSUE #262: bug: `db:dump` crashes trying to serialize `bigint` values
**Describe the bug**

After populating the database and trying to dump the data, the script crashes.

**To Reproduce**

1. Run `deno task db:seed` in a newly created `SaasKit` project.
2. Run `deno task db:dump`, the script will crash giving a JSON serialization error.

**Expected behavior**

The script should output the dumped data.

**Screenshots**

image

**Desktop:**

- OS: MacOS Ventura (13.4)
- Browser N/A
- Version N/A

Jabolol:
Fixes #262 by adding an extensible `replacer` function that handles `BigInt` values correctly

### DIFF
--- a/tools/dump_kv.ts
+++ b/tools/dump_kv.ts
@@ -3,27 +3,17 @@
 // Usage: deno run -A --unstable tools/dump_kv.ts
 import { kv } from "@/utils/db.ts";
 
-interface ToReplace {
-  bigint: bigint;
+// https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-521460510
+function replacer(_key: unknown, value: unknown) {
+  return typeof value === "bigint" ? value.toString() : value; // return everything else unchanged
 }
-
-function applyReplacer<T extends keyof ToReplace>(value: ToReplace[T]) {
-  const revivers: { [K in keyof ToReplace]?: (value: ToReplace[K]) => string } =
-    {
-      bigint: (value) => value.toString(),
-    };
-  return revivers[typeof value as T]?.(value) ?? value;
-}
-
 export async function dumpKv() {
   const iter = kv.list({ prefix: [] });
   const items = [];
   for await (const res of iter) {
     items.push({ [res.key.toString()]: res.value });
   }
-  console.log(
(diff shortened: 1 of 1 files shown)


tools/dump_kv.ts
```diff
@@ -3,13 +3,27 @@
...
+}
+
+function applyReplacer<T extends keyof ToReplace>(value: ToReplace[T]) {
+  const revivers: { [K in keyof ToReplace]?: (value: ToReplace[K]) => string } =
+    {
+      bigint: (value) => value.toString(),
+    };
+  return revivers[typeof value as T]?.(value) ?? value;
+}
+
```
iuioiua, line 17:
```suggestion
// https://github.com/GoogleChromeLabs/jsbi/issues/30#issuecomment-521460510
function replacer(_key: unknown, value: unknown) {
  return typeof value === "bigint" ? value.toString() : value; // return everything else unchanged
}
```
This has the same effect and is more straightforward 🙂

tools/dump_kv.ts
```diff
@@ -3,13 +3,27 @@
...
 export async function dumpKv() {
   const iter = kv.list({ prefix: [] });
   const items = [];
   for await (const res of iter) {
     items.push({ [res.key.toString()]: res.value });
   }
-  console.log(`${JSON.stringify(items, null, 2)}`);
+  console.log(
+    `${JSON.stringify(items, (_key, value) => applyReplacer(value), 2)}`,
+  );
```
iuioiua, line 26:
```suggestion
  console.log(`${JSON.stringify(items, replacer, 2)}`);
```
Ditto

iuioiua (changes requested):
Nice catch! I've suggested a simpler solution.

Jabolol:
> Nice catch! I've suggested a simpler solution.

Yup, simpler one looks great. I created one ready for more `replaces`, in case it was needed, but I highly doubt it would be needed.

iuioiua (approved):
Suggestions applied. Thank you, @Jabolol!

Issue #261, opened 2023-06-17
TypeError: Cannot read properties of null (reading 'unit_amount')


nicu-chiciuc:
**Describe the bug**
When clicking the `Pricing` button on the homepage I get the `TypeError: Cannot read properties of null (reading 'unit_amount')` in `pricing.tsx`.

**To Reproduce**
0. Clone and setup based on Readme.
1. Go to http://localhost:8000/pricing
2. See error on screen, also see error in terminal:
```
An error occurred during route handling or page rendering. TypeError: Cannot read properties of null (reading 'unit_amount')
    at comparePrices (file:///Users/nicu/dev/my/legis_gpt/saaskit/routes/pricing.tsx:18:52)
    at Array.sort (<anonymous>)
    at GET (file:///Users/nicu/dev/my/legis_gpt/saaskit/routes/pricing.tsx:28:27)
    at eventLoopTick (ext:core/01_core.js:183:11)
    at async handler (file:///Users/nicu/dev/my/legis_gpt/saaskit/routes/_middleware.ts:29:15)
    at async Server.handler (https://deno.land/x/fresh@1.1.6/src/server/context.ts:334:19)
    at async Server.#respond (https://deno.land/std@0.178.0/http/server.ts:299:18)
```

**Expected behavior**
I don't expect an error.

**Screenshots**
<img width="1792" alt="Screenshot 2023-06-17 at 15 01 35" src="https://github.com/denoland/saaskit/assets/4076804/c17c886d-739a-4c34-b50f-17e4089a9afc">

**Desktop**
- OS: Apple M1 Pro
- Browser Chrome
- Version 114.0.5735.133 (Official Build) (arm64)

**Additional context** 
I didn't look too much into it, but I didn't expect to get an `any` from the `strip.products.list`, which I assume causes the error.

![Screenshot 2023-06-17 at 15 04 16](https://github.com/denoland/saaskit/assets/4076804/dfed71bc-0799-4fb6-a35f-c4bcc703641b)

iuioiua:
Hi @nicu-chiciuc, I see that you followed the setup in the README. Specifically, did you complete the actions listed in the [Payments and Subscriptions (Stripe)](https://github.com/denoland/saaskit/blob/main/README.md#payments-and-subscriptions-stripe) section?

iuioiua:
[...]

PR #264, opened 2023-06-17
fix: Stripe types and checks

nicu-chiciuc:
Fixes https://github.com/denoland/saaskit/issues/261

## Adding types

I started by adding
```
// @deno-types="npm:stripe"
```
just above the `Stripe` import.
This is based on https://deno.com/manual@v1.34.0/advanced/typescript/types#providing-types-when-importing

This is not the best solution, but currently there's an issue with having types for Stripe in Deno: 
https://github.com/stripe-samples/stripe-node-deno-samples/issues/2

## Covering all possibilities
After the initial error was fixed, I was receiving a different error, also caused by type assertions, so I had to change
```
pricePerInterval={toPricePerInterval(
                product.default_price as Stripe.Price,
              )}
```
and handle all cases

## Testing
1. `git clone ...`
2. `deno task start`
3. access http://localhost:8000/pricing
4. you shouldn't see a 500 error

          
Screenshot 2023-06-18 at 02 28 30

## PS
`defaultPrice.recurring?.interval` already has a defined type, but I wasn't sure what's the correct way to import it in Deno. I left it as string since it's not a critical part of the app, it's used just for displaying text. So it's not a high-concern at the moment.

Screenshot 2023-06-18 at 02 29 37

### DIFF
--- a/utils/payments.ts
+++ b/utils/payments.ts
@@ -26,7 +26,7 @@ if (stripe) {
 
 /**
  * We assume that the product has a default price.
- * The offical types allow for the default_price to be `undefined | null | string`
+ * The official types allow for the default_price to be `undefined | null | string`
  */
 export type StripProductWithPrice = Stripe.Product & {
   default_price: Stripe.Price;
@@ -37,9 +37,8 @@ export function isProductWithPrice(
 ): product is StripProductWithPrice {
   return product.default_price !== undefined &&
     product.default_price !== null &&
-    typeof product.default_price !== "string"
+    typeof product.default_price !== "string";
 }
-
 
 export function formatAmountForDisplay(
   amount: number,
--- a/utils/payments.ts
+++ b/utils/payments.ts
@@ -1,7 +1,4 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
-//
-// Default types for Stripe don't yet work: https://github.com/stripe-samples/stripe-node-deno-samples/issues/2
-// @deno-types="npm:stripe"
 import Stripe from "stripe";
 
 const STRIPE_SECRET_KEY = Deno.env.get("STRIPE_SECRET_KEY");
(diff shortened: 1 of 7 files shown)


nicu-chiciuc:
What should https://github.com/denoland/saaskit/pull/264/commits/cd476cdb302e1b9d35b8fd26c504637dd1e6a423 accomplish?

utils/payments.ts
```diff
@@ -1,4 +1,7 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
+
+// Default types for Stripe don't yet work: https://github.com/stripe-samples/stripe-node-deno-samples/issues/2
+// @deno-types="npm:stripe"
```
iuioiua, line 4: Can you move this to a `stripe.ts` file at the root of this codebase and point the `stripe` import in the `import_map.json` to that file? That way, all files that import `stripe` also get types.

Also, can you please add a version that ensures the types and runtime versions of `stripe` match?

iuioiua (changes requested):
My main thought is that the checks added in this PR aren't needed if the `init:stripe` task is run correctly. So having `product.default_price` as `Stripe.Price` is acceptable. If `product.default_price` is not of type `Stripe.Price`, we may have to investigate the `init:stripe` task, as it's not working correctly.

Either way, nice work fixing the Stripe types! I've added a suggestion to improve this situation for the whole codebase.

nicu-chiciuc:
I see, in that case a better solution would've been for me to not make these checks but throw a custom error/banner instead of relying on type assertions. That means I could remove most of the code checks.

iuioiua:
Yes, I'd rather the code threw 👍🏾

nicu-chiciuc:
I'm made the changes. Remove the type assertions and threw and error (had to create a type guard for this to work).

I still get an error on the pricing page, even after running `deno task init:stripe` since one of the elements in `data` doesn't have a `default_price`. But that's consistent with the previous implementation.

nicu-chiciuc:
I've also moved the stripe definitions into `stripe.ts`. I had to change import in `stripe-webhooks.ts` so it uses a default export like all the other files.

iuioiua (approved):
I made a few fixes, and now LGTM! Thanks for this!

Note for future: please use `deno task ok` before `git push`-ing to ensure all checks pass.

Issue #260, opened 2023-06-17
todo: bump fresh to 1.2.0


huai-jie:
I saw fresh 1.2.0 has been released! Let's try the new features of it! 🤩

iuioiua:
Agreed! It'll be cool to see all the new features of Fresh being taken advantage of! References for those wanting to implement this:
* https://deno.com/blog/fresh-1.2
* https://github.com/denoland/fresh/compare/1.1.6...1.2.0

iuioiua:
This was completed in #274.

PR #274, opened 2023-06-20
chore: update to `fresh@1.2.0`

brunocorrea23:
#260 

For some reason when I updated fresh version to 1.2, the button on the island VoteButton.tsx stopped working. I had to remove the disabled={!IS_BROWSER}. Doesn't seem to have any impact.


iuioiua (changes requested):
Thanks for this. Can you please run the upgrade script using `deno run -A -r https://fresh.deno.dev/update .`? We also want to ensure we take advantage of all the new features. Some ones that we should probably include are:
1. https://github.com/denoland/fresh/pull/1229/files
2. https://github.com/denoland/fresh/pull/757

brunocorrea23:
makes sense. let me investigate these two new features and try to use them.

iuioiua:
Cool. Hopefully, adding these in makes things easier/better. If that's not the case, feel free to leave them out.

brunocorrea23:
@iuioiua , can you check if this last commit makes sense?
I basically created an _app.tsx wrapper, Layout component to it and removed the Layout component from the specific routes.

For the feature that allows sharing signals between islands: not sure if this is useful here so far. We only have the VoteButton.tsx island and I couldn't see much advantage in having signals being passed via props there. What do you think?

components/PageSelector.tsx
```diff
@@ -17,7 +17,7 @@ export default function PageSelector(
           max={props.lastPage}
           value={props.currentPage}
           // @ts-ignore: this is valid HTML
-          onchange="this.form.submit()"
+          onchange={(e) => this.form.submit()}
```
iuioiua, line 20:
```suggestion
          onchange={() => this.form.submit()}
```
If the event parameter isn't used, it can be omitted.

routes/_app.tsx
iuioiua, line 1: Niiiiiice

twind.config.ts
```diff
@@ -1,6 +1,6 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import { Options } from "$fresh/plugins/twindv1.ts";
-import { defineConfig, Preset } from "twind";
+import { defineConfig, Preset } from "https://esm.sh/@twind/core@1.1.3";
```
iuioiua, line 3: Can you please revert? I don't think this is needed.

iuioiua (changes requested):
Oh, this is looking nice. Thoughts:
1. `import Layout from "@/components/Layout.tsx";` can be removed from all routes, except `_app.tsx`.
2. Having `ctx.state` in most instances of `ctx.render()` is no longer needed. This will need to be validated on a case-by-case basis.

twind.config.ts
brunocorrea23: Can you please confirm this here?
I needed to change this because once I upgraded fresh to 1.2, defineConfig and Preset were not defined in "twind" anymore.
This guide here (https://fresh.deno.dev/docs/examples/using-twind-v1) also imports defineConfig from core@1.1.3

brunocorrea23:
> 2. Having `ctx.state` in most instances of `ctx.render()` is no longer needed. This will need to be validated on a case-by-case basis.

If we remove the ctx.state, the sessionId doesn't get passed to the _app.tsx and the Layout component doesn't change as it is supposed to change (when the user is logged in, show 'account' in the navbar)

iuioiua (approved):
I've made a couple of tweaks, and now, LGTM! I'll land merge as-is now. I've asked whether creating the signal within the component or island is best. I'll make an issue if we should proceed with that or not.

Thanks for this, @brunocorrea23!

PR #277, opened 2023-06-21
chore: resize images / improve web perf

brunocorrea23:
#268 

- Resizes github avatar on UserPostedAt.tsx. (main performance opportunity on lighthouse)
- Resizes Logo.webp (performance opportunity on lighthouse)
- Adds padding to vote button to increase tap area (SEO opportunity on lighthouse)

PR #278, opened 2023-06-21
fix: avatar size

brunocorrea23:
after the previous PR #277 , lighthouse started complaining about low resolution of the avatar images. I'm increasing the size from 24 to 36 and it should solve this issue.

Screenshot 2023-06-21 at 02 42 22


iuioiua (approved):
Yeah, I saw that too 😄 LGTM

PR #279, opened 2023-06-21
fix: add reference to unstable types

smitssjors:
As Deno KV is still unstable, its types are not included in `deno.ns`. This means that when opening the project in VS Code, it does not recognise functions such as `Deno.openKv()` (regardless whether the `"deno.unstable": true` VS Code option is set). By referencing `deno.unstable` we also include these types.


iuioiua:
Thanks, @smitssjors. I don't see the same issue in my VS Code. Can you share exactly where you're seeing this?

smitssjors:
I freshly cloned the project and it is actually not showing up now but I am not sure why.

I first ran into this issue when creating a new fresh project. I figured as SaaSKit also uses KV it had some configuration which fixes it but upon cloning I ran into the same issue. After some looking around I did not find any relevant configuration differences as far as I know.

I also just created a new fresh project and the issue is still there.

This might be some weird stuff with VS and/or the plugin and/or deno lsp so im not sure where to look but adding the reference to the fresh project did fix it.

`deno --version`
```sh
deno 1.34.3 (release, x86_64-apple-darwin)
v8 11.5.150.2
typescript 5.0.4
```

VS Code
```
Version: 1.79.2
Commit: 695af097c7bd098fbf017ce3ac85e09bbc5dda06
Date: 2023-06-14T08:58:52.392Z
Electron: 22.5.7
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin x64 22.5.0
```

Deno VS Code plugin version 3.19.1

iuioiua:
How about we add `"deno.unstable": true` to `.vscode/settings.json`? That should be there anyway. We can revisit this if someone comes across this issue in the future.

smitssjors:
I checked in a new GitHub Codespace and the issue exists there too.
But the fact that you can use the unstable APIs without error even though it is not mentioned anywhere is probably a bug.

As this is not the place to discuss or fix it I'll close the PR and open an issue in the repo for the extension.

You can set `"deno.unstable": true` but for me it did not make a difference.

iuioiua:
Actually, you were right. Let's add this in.

Ref: https://github.com/denoland/deno/issues/19035#issuecomment-1539076450

iuioiua (approved):
LGTM! Thank you.

PR #282, opened 2023-06-22
refactor: cleanup `twind`-related functionality

iuioiua:
Everything appears fine after the changes.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -24,16 +24,12 @@
     "preact-render-to-string": "https://esm.sh/*preact-render-to-string@6.1.0",
     "@preact/signals": "https://esm.sh/*@preact/signals@1.1.3",
     "@preact/signals-core": "https://esm.sh/*@preact/signals-core@1.2.3",
-    "twind": "https://esm.sh/twind@0.16.19",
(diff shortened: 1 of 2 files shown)


iuioiua:
@lino-levan, looks like `presetAutoprefix()` may not be needed (ref: https://fresh.deno.dev/docs/examples/using-twind-v1).

lino-levan:
`presetAutoprefix()` as its name implies is useful for targeting features that still have prefixes. Definitely a few sharp corners if you choose not to use it (especially around webkit) that I've personally encountered. It isn't strictly necessary but I'd still support having it on by default for beginners.

lino-levan:
To follow up on the comment, I wanted to check that I wasn't going insane, but no there are definitely some tailwind features that need auto-prefixing.

This is one of the sites I helped build (with fresh) in Safari
image

This is how it renders in chrome without the autoprefix
image

This is how it renders in chrome with autoprefix
image

iuioiua:
Oh, interesting... I should add a note where `presetAutoprefix()` was that states this fact. Thanks!

lino-levan:
Definitely one of those bugs that are hard to locate. A note for future maintainers would be really nice!

Issue #267, opened 2023-06-19
todo: improve test coverage


iuioiua:
The aim is to have [code coverage](https://app.codecov.io/gh/denoland/saaskit) at 100%.

mbhrznr:
would love to tackle this!

iuioiua:
For those wanting to help, I'd welcome a PR that takes care of test coverage for `Vote` within `utils/db.ts`.

mbhrznr:
since i already familiarized myself w/ `utils` i'd tackle that one as well.

iuioiua:
We're now at almost 100%! Thank you to all that helped out this.

iuioiua:
Re-opening this as coverage has slipped a little as things have changed.

iuioiua:
I've expanded the coverage surface area in #495.

iuioiua:
While not at 100%, test coverage is more than sufficient.

PR #284, opened 2023-06-22
refactor: items logic in KV

iuioiua:
Before, functions within `util/db.ts` did too much. E.g. `createItem()` generated an item, set it in KV, and then returned the generated item. Now, functions only do one thing, making things easier to understand and maintain.

This is part 1 of a series of PRs that simplifies DB/KV logic. This part focuses on items specifically.

Towards #267

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -49,11 +49,11 @@ export function newItemProps(): Omit<Item, "userId" | "title" | "url"> {
 }
 
 /**
- * Sets an item in KV.
+ * Creates a new item in KV. Throws if the item already exists in one of the indexes.
  *
  * @example New item creation
  * ```ts
- * import { newItemProps, setItem, incrementAnalyticsMetricPerDay } from "@/utils/db.ts";
+ * import { newItemProps, createItem, incrementAnalyticsMetricPerDay } from "@/utils/db.ts";
  *
  * const item: Item = {
  *   userId: "example-user-id",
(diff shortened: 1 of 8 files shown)


iuioiua:
CC @kit and @brunocorrea23, reviews are welcome, as this PR touches upon functionality that you guys implemented.

utils/db.ts
```diff
@@ -31,75 +30,106 @@ async function getValues<T>(
...
   const res = await kv.atomic()
-    .check({ key: itemKey, versionstamp: null })
-    .check({ key: itemByTime, versionstamp: null })
-    .check({ key: itemsByUserKey, versionstamp: null })
-    .set(itemKey, item)
-    .set(itemByTime, item)
+    .set(itemsKey, item)
+    .set(itemsByTimeKey, item)
     .set(itemsByUserKey, item)
     .commit();
```
huai-jie, line 79: Just wondering why the `.check` is required before but unnecessary now?

utils/db.ts
```diff
@@ -31,75 +30,106 @@ async function getValues<T>(
...
+  const itemsByUserKey = ["items_by_user", item.userId, item.id];
+
+  const res = await kv.atomic()
+    .delete(itemsKey)
+    .delete(itemsByTimeKey)
+    .delete(itemsByUserKey)
+    .commit();
+
+  if (!res.ok) throw new Error(`Failed to set item: ${item}`);
 }
```
huai-jie, line 96: Hmm... is the `deleteItem` feature only intended for testing purposes? It could be used after missing `itemsByTimeKey` from the old items is tackled or the entire KV is reset. It is a pretty helpful feature in my opinion.

huai-jie (commented):
LGTM! Just some NIT

utils/db.ts
iuioiua: Good point. I've added checks and renamed the function to `createItem()` as it's only ever used to create new items.

utils/db.ts
iuioiua: Actually, `deleteItem()` isn't needed in testing or production. So I just removed it.

> It could be used after missing `itemsByTimeKey` from the old items is tackled or the entire KV is reset.

Sorry, what do you mean?

brunocorrea23:
LGTM!

utils/db.ts
huai-jie:
```
  const res = await kv.atomic()
    .delete(itemsKey)
    .delete(itemsByTimeKey)
    .delete(itemsByUserKey)
    .commit();
```

For now, there is no `itemsByTimeKey` key-value in production so it might be causing issues when allowing the user to able to delete the item. 

And It is quite a handy feature for users in my opinion.

Ah nvm it is being removed

PR #289, opened 2023-06-22
fix(stats): populate empty chart axis

### LINKED ISSUE #276: bug: fix /stats charts axis scale
There are currently two issues with the scales for the axis in the /stats charts:

- [ ] y-axis scale should consist of natural numbers only and start from zero.
- [ ] x-axis dates should be sequential. Today it is only showing the dates for which it has data for. All dates within the interval should be shown. If there is no event for that date, simply assign a 'zero' to it.

Screenshot 2023-06-21 at 00 03 32

mbhrznr:
closes #276.

sets the y-axis' ticks to be of natural numbers.
populates the x-axis with sequential dates.

to prevent us from having "holes" in the retrieved dates, a new `getDatesSince` has been introduced.
the returned array can then can be used to query the necessary fields from our analytics stores.

Bildschirm­foto 2023-06-22 um 23 42 37

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -1,4 +1,5 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import { DAY } from "https://deno.land/std@0.192.0/datetime/constants.ts";
 
 const KV_PATH_KEY = "KV_PATH";
 let path = undefined;
@@ -161,7 +162,8 @@ export function newCommentProps(): Pick<Comment, "id" | "createdAt"> {
 export async function createComment(comment: Comment) {
   const commentsByItemKey = ["comments_by_item", comment.itemId, comment.id];
 
-  const res = await kv.atomic()
+  const res = await kv
+    .atomic()
     .check({ key: commentsByItemKey, versionstamp: null })
     .set(commentsByItemKey, comment)
     .commit();
@@ -216,7 +218,8 @@ export async function createVote(vote: Vote) {
     itemsByTimeKey,
     itemsByUserKey,
   ]);
-  const res = await kv.atomic()
+  const res = await kv
+    .atomic()
     .check(itemRes)
     .check(itemsByTimeRes)
     .check(itemsByUserRes)
@@ -261,7 +264,8 @@ export async function deleteVote(vote: Vote) {
     itemsByTimeKey,
     itemsByUserKey,
   ]);
-  const res = await kv.atomic()
+  const res = await kv
+    .atomic()
     .check(itemRes)
     .check(itemsByTimeRes)
     .check(itemsByUserRes)
@@ -383,9 +387,11 @@ export async function getUserByLogin(login: string) {
 
 export async function getUserBySession(sessionId: string) {
   const usersBySessionKey = ["users_by_session", sessionId];
-  return await getValue<User>(usersBySessionKey, {
-    consistency: "eventual",
-  }) ?? await getValue<User>(usersBySessionKey);
+  return (
+    (await getValue<User>(usersBySessionKey, {
+      consistency: "eventual",
+    })) ?? (await getValue<User>(usersBySessionKey))
+  );
 }
 
 export async function getUserByStripeCustomer(stripeCustomerId: string) {
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -28,6 +28,22 @@ async function getValues<T>(
   return values;
 }
 
+async function getMany<T>(
+  keys: Deno.KvKey[],
+): Promise<Deno.KvEntryMaybe<T>[]> {
+  const res: Deno.KvEntryMaybe<T>[] = [];
+  for (const batch of batches(keys, 10)) {
+    res.push(...(await kv.getMany<T[]>(batch)));
+  }
+  return res;
+}
+
+function* batches<T>(array: T[], size: number): Generator<T[], void> {
+  for (let i = 0; i < array.length; i += size) {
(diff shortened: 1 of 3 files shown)


iuioiua (changes requested):
This PR mostly tweaks spacing to adhere to a particular stylistic preference. Can you revert these changes so the PR only targets the main objective and keeps the diff trim?

mbhrznr:
sorry about the initial formatting mess, `prettier` was still active.

iuioiua (changes requested):

utils/db.ts
```diff
@@ -423,14 +424,25 @@ export async function getAnalyticsMetricsPerDay(
...
+    if (metricsMap.has(String(res.key[1]))) {
+      metricsMap.set(String(res.key[1]), Number(res.value));
+    }
   }
-  return { metricsValue, dates };
+  const dates = Array.from(metricsMap.keys());
+  const metricsValue = Array.from(metricsMap.values());
+
+  return { dates, metricsValue };
 }
```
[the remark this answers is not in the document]
mbhrznr: thanks for the feedback.  
there was indeed quite a lot going on there.

i refactored the logic to be more aligned w/ existing functions which where expecting a `msAgo` argument.  
i also introduced some small helpers:  
- `getDatesSince` to get all dates since a given number of milliseconds ago
- `formatDate` to have a unified way of formatting dates

with the newly created tests it should be more predictable to see what's going on.  
however i'd still await the changes introduced in #296.

iuioiua:
#296 has been merged. PTAL, @mbhrznr, as it has some overlap with this PR.

mbhrznr:
@iuioiua thanks for the heads-up!
pr has been updated accordingly.

@brunocorrea23 great job on #296! 👏 👏

mbhrznr:
following on the discussions of #296:  
would it make sense to refactor our current `getAll[X]CountsByDay` functions to a `getAll[X]CountsSince` variant?
if we'd refactor `/stats` to make use of the current implementation we'd still have the issue that the chart axis is (potentially) populated in a non-sequential manner as there might be dates w/o any value.

iuioiua:
Interesting. My current issue with stats and some of DB is that I think the functions do too much, respectively, though we're slowly getting better. E.g. I'd prefer hardcoding the respective metric instead of having a `metric` parameter in `getAnalyticsMetricListPerDay()`, and having a greater number of smaller, simpler functions.

Regarding your suggestion, what if we did this, using `getDatesSince()`?

Before:
```ts
export async function getAllVisitsCountByDay(options?: Deno.KvListOptions) {
  const iter = await kv.list<bigint>({ prefix: ["visits_count"] }, options);
  const visits = [];
  const dates = [];
    visits.push(Number(res.value));
    dates.push(String(res.key[1]));
  }
  return { visits, dates };
}

const visitsCount = await getAllVisitsCountByDay();
```

After:
```ts
export async function getManyVisitsCounts(dates: string[]) {
  const keys = dates.map((date) => ["visits_count", date]);
  const res = await kv.getMany<bigint[]>(keys);
  return res.map((value) => value);
}

const dates = getDatesSince(DAY * 2);
const visitsCount = await getManyVisitsCounts(dates);
```

This seems more streamlined, IMO.

iuioiua:
I'd be happy to make the above simplifications either to this or another PR to help you out, if you'd like.

mbhrznr:
thanks for the recommendation, i really like that one!
i'd incorporate the changes in couple hours from now

iuioiua:
#312 has been merged. Please feel free to carry on with this PR now.

mbhrznr:
@iuioiua thanks for the most recent changes!  

bumped the limit for the metrics back to 30, while also adding a small `getMany` helper which introduces batching.  
most of the initial tests became invalid or were introduced via other prs, which is why i only added two sample cases to test for `getMany`'s length.

utils/db.ts
```diff
@@ -29,6 +28,36 @@ async function getValues<T>(
   return values;
 }
 
+async function getMany<T>(
```
iuioiua, line 31:
```suggestion
async function getManyValues<T>(
```
To align with the other two KV-helper functions. We'd have to return an array of results instead of entries for this. It'd also be good to have a small explainer in regard to the batching functionality.

utils/db.ts
```diff
@@ -29,6 +28,36 @@ async function getValues<T>(
...
+  keys: Deno.KvKey[],
+): Promise<Deno.KvEntryMaybe<T>[]> {
+  const res: Deno.KvEntryMaybe<T>[] = [];
+  for (const batch of batches(keys, 10)) {
+    res.push(...(await kv.getMany<T[]>(batch)));
+  }
+  return res;
+}
+
+function* batches<T>(array: T[], size: number): Generator<T[], void> {
```
iuioiua, line 41: Have you checked whether the Standard Library has something like this in either `std/collections` or `std/async` that we could use instead? If not, all good.

iuioiua (changes requested):
Mostly LGTM! Few nits.

utils/db.ts
mbhrznr: valid point! adjusted the functionality to match the other helpers.

utils/db.ts
mbhrznr: nice, didn't knew about `std/collections`!
refactored the logic to make use of `chunk`.

iuioiua (approved):
I made a few tweaks, and now, LGTM! Thank you for this! Much appreciated.

PR #295, opened 2023-06-23
fix: default timeAgo

brunocorrea23:
After the refactor, the default timeAgo selector is set for all time.
This PR changes it back to the week being the default.


iuioiua (approved):
Good catch! LGTM. Thank you.

Issue #294, opened 2023-06-23
todo: integrate analytic data into atomic operations


iuioiua:
Right now, separate functions exist to increase analytic metrics. E.g. `incrementAnalyticsMetricPerDay("users_count", new Date());` goes with `createUser()`. It should instead be part of the atomic operation within `createUser()`. This happens in a few places within `utils/db.ts`.

iuioiua:
Completed in #296.

PR #296, opened 2023-06-23
refactor: analytics into atomic operations

brunocorrea23:
targets #294

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -29,6 +29,11 @@ async function getValues<T>(
   return values;
 }
 
+/** Converts `Date` to ISO format that is zero UTC offset */
+export function formatDate(date: Date) {
+  return date.toISOString().split("T")[0];
+}
+
 // Item
 export interface Item {
   userId: string;
@@ -69,6 +74,7 @@ export async function createItem(item: Item) {
   const itemsKey = ["items", item.id];
   const itemsByTimeKey = ["items_by_time", item.createdAt.getTime(), item.id];
   const itemsByUserKey = ["items_by_user", item.userId, item.id];
+  const itemsCountKey = ["items_count", `${formatDate(new Date())}`];
 
   const res = await kv.atomic()
     .check({ key: itemsKey, versionstamp: null })
@@ -77,7 +83,7 @@ export async function createItem(item: Item) {
     .set(itemsKey, item)
     .set(itemsByTimeKey, item)
     .set(itemsByUserKey, item)
-    .sum(analyticsKey("items_count"), 1n)
+    .sum(itemsCountKey, 1n)
     .commit();
 
   if (!res.ok) throw new Error(`Failed to create item: ${item}`);
@@ -179,6 +185,7 @@ export async function createVote(vote: Vote) {
     vote.item.id,
     vote.user.id,
   ];
+  const votesCountKey = ["votes_count", `${formatDate(new Date())}`];
 
   const [itemRes, itemsByTimeRes, itemsByUserRes] = await kv.getMany([
     itemKey,
@@ -196,7 +203,7 @@ export async function createVote(vote: Vote) {
     .set(itemsByUserKey, vote.item)
     .set(votedItemsByUserKey, vote.item)
     .set(votedUsersByItemKey, vote.user)
-    .sum(analyticsKey("votes_count"), 1n)
+    .sum(votesCountKey, 1n)
     .commit();
 
   if (!res.ok) throw new Error(`Failed to set vote: ${vote}`);
@@ -286,6 +293,7 @@ export async function createUser(user: User) {
   const usersKey = ["users", user.id];
   const usersByLoginKey = ["users_by_login", user.login];
   const usersBySessionKey = ["users_by_session", user.sessionId];
+  const usersCountKey = ["users_count", `${formatDate(new Date())}`];
 
   const atomicOp = kv.atomic();
 
@@ -306,7 +314,7 @@ export async function createUser(user: User) {
     .set(usersKey, user)
     .set(usersByLoginKey, user)
     .set(usersBySessionKey, user)
-    .sum(analyticsKey("users_count"), 1n)
+    .sum(usersCountKey, 1n)
     .commit();
 
   if (!res.ok) throw new Error(`Failed to create user: ${user}`);
@@ -405,16 +413,6 @@ export async function getAnalyticsMetricPerDay(metric: string, date: Date) {
     metric,
     `${date.toISOString().split("T")[0]}`,
   ]);
-}
-
-export function analyticsKey(
-  metric: string,
-) {
-  // convert to ISO format that is zero UTC offset
-  return [
-    metric,
-    `${new Date().toISOString().split("T")[0]}`,
-  ];
 }
 
 export async function getAnalyticsMetricListPerDay(
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -63,7 +63,6 @@ export function newItemProps(): Pick<Item, "id" | "score" | "createdAt"> {
  * };
  *
  * await createItem(item);
- * await incrementAnalyticsMetricPerDay("items_count", item.createdAt);
  * ```
  */
 export async function createItem(item: Item) {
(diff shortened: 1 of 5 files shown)


iuioiua (changes requested):
Great! Thanks for this. Please also check that their respective functions have increased the analytics data. E.g. that `users_count` has been incremented in the `createUser()` test.

Can we also investigate removing `incrementAnalyticsMetricPerDay()`?

brunocorrea23:
Thanks!
- I have added tests for createUser() and createItem(). I didn't do it for createVote() since this test has not been created yet and is part of an ongoing PR. We can add that later.
- I've removed the incrementVisitsPerDay() in favor of a more generic incrementAnalyticsMetricPerDay() that can be reused if more cases in the future. Also did some refactoring around that. Let me know if this makes sense. I can undo these changes if it doens't.

utils/db.ts
```diff
@@ -401,25 +400,24 @@ export async function incrementAnalyticsMetricPerDay(
...
-  // convert to ISO format that is zero UTC offset
-  const visitsKey = [
-    "visits",
-    `${date.toISOString().split("T")[0]}`,
-  ];
-  await kv.atomic()
-    .sum(visitsKey, 1n)
-    .commit();
-}
-
```
iuioiua, line 414: Niiiice

utils/db.ts
```diff
@@ -308,6 +306,7 @@ export async function createUser(user: User) {
     .set(usersKey, user)
     .set(usersByLoginKey, user)
     .set(usersBySessionKey, user)
+    .sum(analyticsKey("users_count"), 1n)
```
iuioiua, line 309: Hm... It might be better to define a `const usersCount` at the start of the function with the others without using `analyticsKey()`. It gives a more immediate idea of what's going on, IMO. At most, maybe a `toDate()` that returns `date.toISOString().split("T")[0]` might be useful. Ditto for other similar scenarios. WDYT?

utils/db_test.ts
```diff
@@ -71,7 +72,14 @@ Deno.test("[db] (get/create)Item()", async () => {
 
   assertEquals(await getItem(item.id), null);
 
+  const itemsCount =
+    (await getAnalyticsMetricPerDay("items_count", new Date()))?.valueOf() ??
+      0n;
```
iuioiua, line 77:
```suggestion
  assertEquals(await getAnalyticsMetricPerDay("items_count", new Date()), null);
```

iuioiua (changes requested):
IMO, we should drop `getAnalyticsMetricPerDay()` in favour of, say `getItemsCountByDate()`, and the same for the others. I'd rather have more lines of code that are easier to understand rather than fewer lines of code that require more thinking. WDYT?

utils/db_test.ts
```diff
@@ -71,7 +72,14 @@ Deno.test("[db] (get/create)Item()", async () => {
...
   await createItem(item);
+  assertEquals(
+    (await getAnalyticsMetricPerDay("items_count", new Date()))!.valueOf(),
```
iuioiua, line 80: Is `.valueOf()` needed? Ditto elsewhere.

iuioiua (changes requested):

utils/db.ts
iuioiua: I.e. check out `formatDate()` in #289.

utils/db_test.ts
brunocorrea23: It seems so. This is the test result when I remove the velueOf():

[Diff] Actual / Expected
-   [Deno.KvU64: 3n]
+   3n

utils/db_test.ts
brunocorrea23: @iuioiua , can you pls explain this suggestion here?
I believe that if you do it like this, the assertion is going to fail. When it gets to this point, the itemsCount is probably equal to `2n` because of the two `createItem()` that were run on the test above (`" [db] getAllItems()"`)

We can't assume anything about the value o itemsCount, as we don't know how many tests will use it before in the code. We can only test that the itemsCount_after = itemsCount_before + 1.

Does it make sense? Did I understand correctly your suggestion?

utils/db.ts
brunocorrea23: Makes sense. Will adjust that!

brunocorrea23:
> IMO, we should drop `getAnalyticsMetricPerDay()` in favour of, say `getItemsCountByDate()`, and the same for the others. I'd rather have more lines of code that are easier to understand rather than fewer lines of code that require more thinking. WDYT?

Thanks for your input here 🙏  I'll do some refactoring around this so that we end up with 3 functions per metric:
- increment[X]ByDay(date: Date)
- get[X]ByDay(date: Date)
- get[X]ListByDay(options?: Deno.KvListOptions)

utils/db_test.ts
iuioiua: I see. Fair enough.

utils/db_test.ts
iuioiua: Perhaps, we should move the line that defines `itemsCount` elsewhere such that it doesn't require a default value. WDYT?

utils/db.ts
```diff
@@ -69,6 +74,7 @@ export async function createItem(item: Item) {
   const itemsKey = ["items", item.id];
   const itemsByTimeKey = ["items_by_time", item.createdAt.getTime(), item.id];
   const itemsByUserKey = ["items_by_user", item.userId, item.id];
+  const itemsCountKey = ["items_count", `${formatDate(new Date())}`];
```
iuioiua, line 77:
```suggestion
  const itemsCountKey = ["items_count", formatDate(new Date())];
```

utils/db.ts
```diff
@@ -179,6 +185,7 @@ export async function createVote(vote: Vote) {
     vote.item.id,
     vote.user.id,
   ];
+  const votesCountKey = ["votes_count", `${formatDate(new Date())}`];
```
iuioiua, line 188:
```suggestion
  const votesCountKey = ["votes_count", formatDate(new Date())];
```

brunocorrea23:
@iuioiua , couple of comments here:
- I've created the analytics functions I've mentioned above.
- I haven't changed the way /stats uses the analytics metrics as I believe this is outside the scope of this PR and might overlap with the work on #289 . If the changes are not handled by that PR, I can create another one soon after it.
- As for the default value on the db_test.ts you've mentioned above: the way I see to remove the need for a default value is simply to move the function to the top of the file, where we can guarantee that no one has increased that metrics count before. But I do think that the current implementation is more robust, as it doesn't assume anything about order of functions within that file. I can imagine in the future someone trying to changing that file, the test starts failing and it will be hard to debug why.

utils/db.ts
```diff
@@ -387,39 +394,90 @@ export function compareScore(a: Item, b: Item) {
 }
 
 // Analytics
-export async function incrementAnalyticsMetricPerDay(
-  metric: string,
-  date: Date,
-) {
+export async function incrementVisitsCountByDay(date: Date) {
```
iuioiua, line 397:
```suggestion
export async function incrVisitsCountByDay(date: Date) {
```
Nit. WDYT? Is it still clear to a newcomer what this does?

utils/db.ts
```diff
@@ -387,39 +394,90 @@ export function compareScore(a: Item, b: Item) {
...
+    "items_count",
     `${date.toISOString().split("T")[0]}`,
   ]);
 }
 
-export async function getAnalyticsMetricsPerDay(
+export async function getVotesCountByDay(date: Date) {
+  return await getValue<bigint>([
+    "votes_count",
+    `${date.toISOString().split("T")[0]}`,
```
iuioiua, line 425:
```suggestion
    formatDate(date),
```

utils/db.ts
```diff
@@ -387,39 +394,90 @@ export function compareScore(a: Item, b: Item) {
...
+  ]);
+}
+
+export async function getUsersCountDay(date: Date) {
+  return await getValue<bigint>([
+    "users_count",
+    `${date.toISOString().split("T")[0]}`,
```
iuioiua, line 432:
```suggestion
    formatDate(date),
```

iuioiua (changes requested):
Few nits

utils/db.ts
brunocorrea23: Makes sense. I think it is still clear!

brunocorrea23:
Addressed the previous comments. Please take a look.

iuioiua (approved):
Great work! Nice cleanups and simplifications! Thanks for this one, @brunocorrea23.

iuioiua:
Closes #294

brunocorrea23:
@iuioiua , thank you for all the help and input! 🙏

PR #299, opened 2023-06-25
fix(blog): respect user theme choice

niklasmtj:
At the moment the blog is not readable since it has css classes set only for a light theme. Since SaaSKit changed to support both light and dark themes the blog also needs to support that. This fixes the CSS on the SaaSKit side. However since we use `deno-gfm` we might think about adding our own CSS to for the rendered markdown files.


lino-levan:
`deno-gfm` supports dark mode if properly configured.

iuioiua (approved):
LGTM! Thank you for this. I've created #301 to address the dark theme styling of markdown files.

PR #303, opened 2023-06-26
test(utils): further improve test coverage

mbhrznr:
this pr attempts to improve the coverage as mentioned in #267.  

to reduce the boilerplate on writing further tests, a couple `genNew<Entity>` functions have been created.  
these helper functions are now used for all tests cases, if applicable.

### DIFF
--- a/data/posts/second-post.md
+++ b/data/posts/second-post.md
@@ -1,7 +1,5 @@
 ---
 title: Second post
-published_at: 2022-03-20
-summary: This is the second post.
 ---
 
 It was popularised in the 1960s with the release of Letraset sheets containing
--- a/routes/blog/[slug].tsx
+++ b/routes/blog/[slug].tsx
@@ -20,9 +20,10 @@ export const handler: Handlers<BlogPostPageData, State> = {
 
 export default function PostPage(props: PageProps<BlogPostPageData>) {
   const { post } = props.data;
-  const date = new Date(post.publishedAt).toLocaleDateString("en-US", {
-    dateStyle: "long",
-  });
+  const date = post.publishedAt.toString() !== "Invalid Date" &&
+    new Date(post.publishedAt).toLocaleDateString("en-US", {
+      dateStyle: "long",
+    });
 
   return (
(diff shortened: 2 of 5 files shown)


data/posts/second-post.md
```diff
@@ -1,7 +1,5 @@
 ---
 title: Second post
-published_at: 2022-03-20
```
mbhrznr, line 3: we were missing just a tiny bit on the [coverage](https://app.codecov.io/gh/denoland/saaskit/blob/main/utils%2Fposts.ts) for `posts`, so i wanted to incorporate an example, which would check the fallbacks for `published_at` and `summary`.

routes/blog/[slug].tsx
```diff
@@ -20,9 +20,10 @@ export const handler: Handlers<BlogPostPageData, State> = {
 
 export default function PostPage(props: PageProps<BlogPostPageData>) {
   const { post } = props.data;
-  const date = new Date(post.publishedAt).toLocaleDateString("en-US", {
-    dateStyle: "long",
-  });
+  const date = post.publishedAt.toString() !== "Invalid Date" &&
```
Comment (mbhrznr) [inline routes/blog/[slug].tsx:23]: whenever we had an `Invalid Date`, it would get rendered as well.  
this aligns our logic with `/blog/index.tsx`, where we explicitly check for `Invalid Date`.

utils/posts.ts
```diff
@@ -31,7 +31,7 @@ export async function getPost(slug: string): Promise<Post | null> {
     return {
       slug,
       title: attrs.title as string,
-      publishedAt: new Date(attrs.published_at as Date) || null,
```
mbhrznr, line 34: turns out that `|| null` wouldn't work here.  
even though `new Date(undefined)` returns `Invalid Date` it'd still be a truthy value.
Bildschirm­foto 2023-06-26 um 21 37 34

iuioiua (commented):
Great work! LGTM. Thank you 🙏🏾

iuioiua (approved):

Issue #293, opened 2023-06-23
todo: cleanup dummy data from KV


iuioiua:
Dummy data, such as items and comments with "test" or "hello, should be removed in a migration script.

PR #306, opened 2023-06-28
feat: `deleteItem()` and `deleteComment()`

iuioiua:
Towards #293

CC @huai-jie


huai-jie:
Ah, so currently it is used for deleting the dummy data. 

Anyway, do you think allowing users to delete the items and comments they have created can be a valuable feature in this case?

PR #312, opened 2023-06-28
refactor: simplify analytics-KV data

iuioiua:
Previously, the analytics-related functions in `@/utils/db.ts` did too much. This PR simplifies these functions. I also chose not to hardcode the specific metric value and instead it be a function argument.

PTAL, @brunocorrea23 and @mbhrznr.
Pre-requisite for #289

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -437,7 +437,7 @@ export function getDatesSince(msAgo: number) {
 
   while (+start < now) {
     start.setDate(start.getDate() + 1);
-    dates.push(new Date(start));
+    dates.push(formatDate(new Date(start)));
   }
 
   return dates;
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -422,7 +422,7 @@ export async function incrVisitsCountByDay(date: Date) {
   // convert to ISO format that is zero UTC offset
   const visitsKey = [
     "visits_count",
-    `${date.toISOString().split("T")[0]}`,
+    formatDate(date),
   ];
   await kv.atomic()
     .sum(visitsKey, 1n)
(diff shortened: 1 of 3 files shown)


iuioiua:
Actually, it might be best to use `.list()`.

utils/db.ts
```diff
@@ -422,106 +422,32 @@ export async function incrVisitsCountByDay(date: Date) {
...
+  metric: "visits_count" | "items_count" | "votes_count" | "users_count",
+  dates: Date[],
 ) {
-  const analyticsByDay = await Promise.all(
-    metrics.map((metric) => getAnalyticsMetricListPerDay(metric, options)),
-  );
-
-  return analyticsByDay;
+  const keys = dates.map((date) => [metric, formatDate(date)]);
+  const res = await kv.getMany<bigint[]>(keys);
```
mbhrznr, line 451: **praise:** nice to see the refactoring to `.getMany`.  
this was also one of the first approaches i had in mind when starting w/ #289.  

**question:** would it make sense to check `dates.length` so we don't run in any issues w/ the limit?  
if we want to continue that route, would it also make sense to split `dates` into chunks of 10s?

utils/db.ts
```diff
@@ -422,106 +422,32 @@ export async function incrVisitsCountByDay(date: Date) {
...
-  }
-  return { metricsValue, dates };
+  return dates;
 }
 
-export async function getManyAnalyticsMetricsPerDay(
-  metrics: string[],
-  options?: Deno.KvListOptions,
+export async function getManyMetrics(
+  metric: "visits_count" | "items_count" | "votes_count" | "users_count",
```
mbhrznr, line 447: **praise:** loving the union type here!

mbhrznr (approved):
lgtm! well done!

utils/db.ts
iuioiua: `kv.getMany()` will throw if the limit is exceeded anyway, so I think it's fine in its current state. If needed, we can work on "batching" this function up in another PR.

PR #313, opened 2023-06-29
fix: remove deprecated `incrementAnalyticsMetricPerDay()`

boscohyun:
`incrementAnalyticsMetricPerDay` does not exist anymore in `db.ts`.
This pull request just remove invalid imports.
When I ran the `db:seed` task during `saaskit` startup, I got an error, and this pull request was enough to get rid of it.


iuioiua (approved):
Good catch! Thank you.

PR #320, opened 2023-06-29
fix: stats graphs alignments

iuioiua:
This change horizontally aligns graphs to the centre of the page on desktop screens.

This is a follow-up to #319
Depends on https://github.com/denoland/fresh_charts/pull/14

CC @kt3k


kt3k (approved):
LGTM

Issue #228, opened 2023-05-24
proposal: receive notifications when someone comments on or upvotes your item


huai-jie:

huai-jie:
I'm curious about how this could be achieved using Fresh. Are there any available resources for accomplishing this?

iuioiua:
Completed in #321.

PR #321, opened 2023-06-30
feat: notifications

brunocorrea23:
This PR is linked to #228 
Although the goal now is not to develop new features, I was curious to work on this one.

It does:
- adds notification badge to the navbar
- creates a /notifications page
- generates notifications for new comments on a user's post/item
- tests for the new functions on db.ts
- a notification is deleted once the user clicks on it

Next steps I imagine here:
- add notifications for votes
- add pagination for the /notifications page
- manage read/unread status. if the user clicks on a notification, it is marked as read, but not deleted. If the user wants, it can delete the notification in another button (similar to what happens in an email)
- improve design of the notification items
- add a realtime notification on the browser. reference: https://github.com/almahdi/Notifire

Excited to hear your thoughts on it!


iuioiua:
Wow. This is very interesting. I like the idea a lot. Let me test and play around with it.

iuioiua (changes requested):
Very interesting idea. I've given it a quick read-through.

routes/notifications/_middleware.ts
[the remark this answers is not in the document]
brunocorrea23: ops! great catch

components/Layout.tsx
```diff
@@ -83,6 +87,16 @@ export default function Layout(props: LayoutProps) {
...
+    props.session && props.notificationsCount! > 0
+      ? {
+        href: "/notifications",
+        inner: (
+          <span class={notificationsStyle}>
+            {props.notificationsCount}
+          </span>
+        ),
+      }
+      : { href: "/", inner: "" },
```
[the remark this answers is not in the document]
brunocorrea23: much better indeed

utils/db.ts
```diff
@@ -171,6 +171,109 @@ export async function getItemsSince(msAgo: number) {
...
+  return await getValues<Notification>({
+    prefix: ["notifications_by_user", userId],
+  });
+}
+
+export async function getNotificationsCountByUser(userId: string) {
+  const notificationsCountByUser =
+    (await getNotificationsByUser(userId)).length;
+  return notificationsCountByUser;
+}
```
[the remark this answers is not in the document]
brunocorrea23: ok! I've done the changes. But just so I learn: why choosing to do this here?

components/Layout.tsx
```diff
@@ -83,6 +87,16 @@ export default function Layout(props: LayoutProps) {
         href: "/signin",
         inner: "Sign in",
       },
+    props.session && props.notificationsCount! > 0
+      ? {
+        href: "/notifications",
+        inner: (
+          <span class={notificationsStyle}>
+            {props.notificationsCount}
```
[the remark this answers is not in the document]
brunocorrea23: I'm not sure I got this here.
In GitHub, the blue dot is on the /notifications link, not on the accounts link.

In any case, if the dot was in the accounts link, what would happen if a user clicked on this link? would it lead the user to the /account page or the the /notifications page?
It could may be the case that the user clicks on this accounts button and a drop down menu appears where a user can select between the action /account page or the /notifications page. I like this approach, but I'd suggest we tackle this later as it seems more complex.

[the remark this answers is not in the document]
iuioiua: Fair points. Let me refine my suggestion:
1. Have a notifications icon link in the header menu.
2. The link goes to `/account/notifications`.
3. When there are "unread" notifications, there is a dot on the link. When all notifications are "read", there is no dot on the link.

WDYT?

utils/db.ts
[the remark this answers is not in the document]
iuioiua: If we use a dot instead of a count, the exact number of unread notifications matters less. What matters is _if_ there are any unread notifications at all. The eventual consistency mode may return a stale value, but it is faster.

components/Layout.tsx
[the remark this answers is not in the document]
brunocorrea23: I tried following your suggestion. Not sure if I liked the final design, but I guess we can prettify it later

routes/_middleware.ts
```diff
@@ -33,7 +35,13 @@ export async function handler(
 
   await incrVisitsCountByDay(new Date());
 
-  ctx.state.sessionId = await getSessionId(req);
+  const sessionId = await getSessionId(req);
+  ctx.state.sessionId = sessionId;
+
+  if (sessionId) {
+    const user = await getUserBySession(sessionId);
+    ctx.state.notificationsCount = await getNotificationsCountByUser(user!.id);
```
iuioiua, line 43:
```suggestion
    ctx.state.hasNotificiations = await ifUserHasNotifications(user!.id);
```
WDYT?

routes/_middleware.ts
```diff
@@ -4,10 +4,12 @@ import { walk } from "std/fs/walk.ts";
 import { getSessionId } from "kv_oauth";
 import { redirect, setRedirectUrlCookie } from "@/utils/redirect.ts";
 import { Status } from "std/http/http_status.ts";
+import { getNotificationsCountByUser, getUserBySession } from "@/utils/db.ts";
 import { incrVisitsCountByDay } from "@/utils/db.ts";
 
 export interface State {
   sessionId?: string;
+  notificationsCount?: number;
```
iuioiua, line 12:
```suggestion
  hasNotifications?: boolean;
```
WDYT?

routes/_middleware.ts
brunocorrea23: LGTM

routes/_middleware.ts
brunocorrea23: LGTM

utils/db.ts
```diff
@@ -271,10 +271,11 @@ export async function getNotificationsByUser(
   }, options);
 }
 
-export async function getNotificationsCountByUser(userId: string) {
+export async function ifUserHasNotifications(userId: string) {
   const notificationsCountByUser =
     (await getNotificationsByUser(userId, { consistency: "eventual" })).length;
-  return notificationsCountByUser;
+  const hasNotifications = notificationsCountByUser > 0 ? true : false;
+  return hasNotifications;
```
iuioiua, line 278:
```suggestion
  return notificationsCountByUser > 0;
```

routes/item/[id].tsx
```diff
@@ -112,7 +112,7 @@ export const handler: Handlers<ItemPageData, State> = {
     };
     await createNotification(notification);
 
-    return redirect(`/item/${ctx.params.id}`);
+    return redirect(`/`);
```
Comment (iuioiua) [inline routes/item/[id].tsx:115]: Why was this done? Am I missing something?

Comment (brunocorrea23) [reply routes/item/[id].tsx]: Probably just some fat finger on my side. Just fixed it now.

utils/db_test.ts
```diff
@@ -236,3 +243,58 @@ Deno.test("[db] getDatesSince()", () => {
     formatDate(new Date()),
   ]);
 });
+
+//ok
```
iuioiua, line 247: Can you please remove this and the others?

utils/db.ts
```diff
@@ -171,6 +171,112 @@ export async function getItemsSince(msAgo: number) {
...
+ *   userFromLogin: "example-user-from-login"
+ *   originId: "example-origin-id"
+ *   originTitle: "example-origin-title"
+ *   ...newNotificationProps(),
+ * };
+ *
+ * await createNotification(notification);
+ * ```
+ */
+export async function createNotification(notif: Notification) {
```
iuioiua, line 214:
```suggestion
export async function createNotification(notification: Notification) {
```
We can use the whole `notification` variable name since we do it everywhere else.

utils/db.ts
```diff
@@ -171,6 +171,112 @@ export async function getItemsSince(msAgo: number) {
...
+  userId: string;
+  type: string;
+  userFromId: string;
+  userFromLogin: string;
+  originId: string;
+  originTitle: string;
+  // The below properties can be automatically generated upon item creation
+  id: string;
+  createdAt: Date;
+}
```
iuioiua, line 185:
```suggestion
export interface Notification {
  userId: string;
  text: string;
  url: string;
  // The below properties can be automatically generated upon item creation
  id: string;
  createdAt: Date;
}
```
Can we simplify the `Notification` interface? The text for the notification would be generated upon creation, and the notification row in `/account/notifications` would link back to where the notification was generated. E.g. link back to the item where a comment was made. This would make extending notification functionality in the future fairly easy. WDYT?

components/Layout.tsx
```diff
@@ -83,20 +92,38 @@ export default function Layout(props: LayoutProps) {
...
+          {props.hasNotifications!
+            ? (
+              <>
+                Notifications <span class="text-blue-500">•</span>
+              </>
+            )
+            : "Notifications"}
+        </span>
+      ),
+    });
```
iuioiua, line 111: To clarify, I think:
1. The notifications link should be an icon.
2. The link should always be visible, with the blue dot only showing when unread notifications exist.
3. The link should still be visible if the user is signed out. The user would be automatically redirected to `/account/notifications` anyway, thanks to current middleware.

iuioiua (changes requested):
This functionality has a few moving parts, so I'd like to get it reasonably right.

utils/db.ts
brunocorrea23: Makes sense! Working on it!
I'd still leave a `title` or `type` field. Will probably be good for filtering in the future.

iuioiua (approved):
LGTM! Let's watch this feature, as it's new territory. Thanks for this!

brunocorrea23:
@iuioiua , thanks for the inputs. Current solution looks much better than the original.

iuioiua:
Looking very nice.
hunt deno land_account_notifications

PR #330, opened 2023-07-02
perf: improve `ifUserHasNotifications()`

iuioiua:
Please review, @brunocorrea23.


brunocorrea23:
LGTM.
Curiosity: why the change?

iuioiua:
Previously, the check would iterate through the entire list of notifications. Now, it returns true on the first iteration.

It's important that this function incurs the smallest performance hit possible, as it's called upon every request.

Issue #335, opened 2023-07-03
bug: notification red dot went to corner of the screen


brunocorrea23:
probably has to do with some relative/absolute twind config

<img width="449" alt="Screenshot 2023-07-03 at 00 10 37" src="https://github.com/denoland/saaskit/assets/8139585/c93f6989-eddd-4eae-a09b-917eed01ad7b">

iuioiua:
I recently made some front-end changes that would've caused this. My bad.

iuioiua:
Completed in #337.

PR #337, opened 2023-07-03
fix: notification dot position

brunocorrea23:
Targets #335 
I added the relative tag on the Header.tsx page, but could also have added it 'globally' on the LINK_STYLES on the constants.ts file.
WDYT?

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -24,7 +24,7 @@ export default function Header(
           : <a href="/signin" class={LINK_STYLES}>Sign in</a>}
         <a
           href="/account/notifications"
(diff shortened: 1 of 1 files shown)


iuioiua (changes requested):
I think the `relative` class more belongs to `<Bell />`.

brunocorrea23:
I believe the problem is that the `relative` must be set in a parent element to `CircleFilled`.
Since `Bell` is at the same level as `CircleFilled`, it wouldn't work.

iuioiua (approved):
LGTM! Thank you.

PR #338, opened 2023-07-03
feat: add first working draft of Chart.js islands

mbhrznr:
this pr adds a first version of a chartjs island as seen in the draft in `fresh_charts`:
https://github.com/denoland/fresh_charts/pull/15

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -30,7 +30,7 @@
     "feed": "https://esm.sh/feed@4.2.2",
     "kv_oauth": "https://deno.land/x/deno_kv_oauth@v0.2.5/mod.ts",
     "@twind/core": "https://esm.sh/@twind/core@1.1.3",
-    "chart.js": "https://esm.sh/chart.js@4.0.1/auto"
+    "chart.js": "https://esm.sh/stable/chart.js@4.0.1/auto"
   },
   "exclude": [
     "cov/"
--- a/deno.json
+++ b/deno.json
@@ -30,7 +30,7 @@
     "feed": "https://esm.sh/feed@4.2.2",
     "kv_oauth": "https://deno.land/x/deno_kv_oauth@v0.2.5/mod.ts",
     "@twind/core": "https://esm.sh/@twind/core@1.1.3",
-    "chart.js": "https://esm.sh/stable/chart.js@4.0.1/auto"
+    "chart.js": "https://esm.sh/stable/chart.js@4.3.0/auto"
   },
   "exclude": [
     "cov/"
--- a/deno.json
+++ b/deno.json
@@ -30,7 +30,7 @@
     "feed": "https://esm.sh/feed@4.2.2",
     "kv_oauth": "https://deno.land/x/deno_kv_oauth@v0.2.5/mod.ts",
     "@twind/core": "https://esm.sh/@twind/core@1.1.3",
-    "chart.js": "https://esm.sh/stable/chart.js@4.3.0/auto"
+    "chart.js": "https://esm.sh/stable/chart.js@4.0.1/auto"
   },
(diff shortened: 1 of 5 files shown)


deno.json
```diff
@@ -28,9 +28,9 @@
     "std/": "https://deno.land/std@0.188.0/",
     "stripe": "./stripe.ts",
     "feed": "https://esm.sh/feed@4.2.2",
-    "fresh_charts/": "https://deno.land/x/fresh_charts@0.2.2/",
     "kv_oauth": "https://deno.land/x/deno_kv_oauth@v0.2.5/mod.ts",
-    "@twind/core": "https://esm.sh/@twind/core@1.1.3"
+    "@twind/core": "https://esm.sh/@twind/core@1.1.3",
+    "chart.js": "https://esm.sh/chart.js@4.0.1/auto"
```
mbhrznr, line 33: intentionally chose to stick to `4.0.1` for now.  
on `4.3.0` some of the core plugins seem to be broken.  
also the rendering seems to be changed, as the canvas doesn't respect the container's with anymore.  
thus the charts would always be rendered blurry at 300x150 each.

routes/stats.tsx
```diff
@@ -122,17 +75,52 @@ export default function StatsPage(props: PageProps<StatsPageData>) {
   );
 
   return (
-    <main class="flex-1 p-4">
-      <div class="gap-4">
-        {charts.map((chart) => (
+    <main class="flex-1 p-4 grid gap-12 md:grid-cols-2">
```
mbhrznr, line 78: thought it would be a neat idea to have the charts rendered as grid again.

iuioiua: I'd rather each take up one line and be much bigger, but my preference isn't strong.

iuioiua:
On my machine, each chart is still quite small.
image

deno.json
mbhrznr: tried to bump to `4.3.0` again.
however the animations as well as the intersection for the tooltips didn't seem to work either.
i'd try to do some further investigations on why that happens.

routes/stats.tsx
mbhrznr: please let me know if i should remove the grid.

mbhrznr:
> On my machine, each chart is still quite small.

interestingly this was one of the issues that only occured when i tried to bump to latest.
and thanks for the preview deployment, i added a `ResizeObserver` to resolve these pesky resizing issues.

iuioiua (changes requested):
Mostly LGTM! Am I right in saying that the chart now supports all properties because it's being rendered client-side? Also, are there any parts within the chart island that can be removed? We should have it as simple as possible.

islands/LineChart.tsx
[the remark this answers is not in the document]
mbhrznr: refactored `Chart.tsx` to be a generic component.
a chart's `type` is the determining factor for chart's supported options as well the rendered kind.

`title` and `total` are now part of `stats.tsx` again.

mbhrznr:
> Am I right in saying that the chart now supports all properties because it's being rendered client-side?

yes, the island should now accept and support (🤞) all known `Chart.js` properties ootb.

> Also, are there any parts within the chart island that can be removed?

i tried to reduce it to a minimum.
while the `useResizeObserver` hook shouldn't be needed in theory, i'd propose to keep it in, as there are some known issues w/ [responsiveness](https://github.com/chartjs/Chart.js/issues/11005) in version 4.

iuioiua (approved):
LGTM for a first draft! I'm very excited to expand on this - I already have a few ideas. Thank you very much for this 🚀

PR #339, opened 2023-07-03
feat: add dependency update task with `udd`

### LINKED ISSUE #20: suggestion: `deno task update` and automatic updates
An `update` task like [this](https://github.com/Hexagon/bundlee/blob/4858ed39e8220a41779c126b80ec918d43287979/deno.json#L9) would be fantastic. This'd also be great to integrate into a weekly GitHub action like [this](https://github.com/hayd/deno-udd/blob/master/.github/workflows/udd.yml) to keep dependencies up-to-date.

niklasmtj:
This addresses and closes #20. I still keep this in draft mode until we finished the discussions in the respective issue.

### DIFF
--- a/.github/workflows/dependency-update.yml
+++ b/.github/workflows/dependency-update.yml
@@ -1,0 +1,53 @@
+name: udd-update-dependencies
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 3 * * 1"
+
+jobs:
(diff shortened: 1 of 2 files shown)


iuioiua:
I will close this, as we determined that `udd` is likely not the best tool for the job. If you find another better one, feel free to open a PR!

PR #343, opened 2023-07-05
refactor: stats charts improvements and simplifications

iuioiua:
This change:
* Simplifies the `<Chart />` island to only return the `<canvas>` element. Parent styling can be done separately.
* Combines 4 charts into a single, large, interactive chart that can hide/show datasets.
* Adds enables the [`index`](https://www.chartjs.org/docs/latest/samples/tooltip/content.html) interaction mode, which shows the values for all shown datasets for a given date.
* Removes configuration and code which may not be needed.
* Disables `maintainAspectRatio`.

Please review by checking deployment, @mbhrznr.

### DIFF
--- a/islands/Chart.tsx
+++ b/islands/Chart.tsx
@@ -5,7 +5,7 @@ import {
   type ChartType,
   type DefaultDataPoint,
 } from "chart.js";
-import { type MutableRef, useEffect, useRef } from "preact/hooks";
+import { useEffect, useRef } from "preact/hooks";
 import type { JSX } from "preact";
 
 type ChartOptions<
@@ -28,7 +28,6 @@ function useChart<
   Data = DefaultDataPoint<Type>,
   Label = unknown,
 >(options: ChartOptions<Type, Data, Label>) {
-  const containerRef = useRef<HTMLDivElement | null>(null);
   const canvasRef = useRef<HTMLCanvasElement | null>(null);
   const chartRef = useRef<ChartJS<Type, Data, Label> | null>(null);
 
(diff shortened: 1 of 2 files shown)


islands/Chart.tsx
```diff
@@ -84,14 +84,9 @@ export default function Chart<
   }, []);
 
   return (
-    <div
-      {...container}
-      ref={containerRef}
```
mbhrznr, line 89: **suggestion:** as we don't have a container anymore i would make sense to remove, `container` from `ChartProps` and `containerRef` from `useChart`.
and as `useResizeObserver` has no truthy target anymore, we could remove the hook as well.

routes/stats.tsx
```diff
@@ -44,89 +44,71 @@ export const handler: Handlers<StatsPageData, State> = {
...
-                  pointRadius: 0,
-                  cubicInterpolationMode: "monotone",
-                }],
-              }}
-            />
-          </div>
-        );
-      })}
+    <main class="flex-1 p-4 aspect-[2/1] mx-auto relative w-full">
+      <Chart
```
mbhrznr, line 79: **praise:** really liking the huge, joint chart!

routes/stats.tsx
```diff
@@ -44,89 +44,71 @@ export const handler: Handlers<StatsPageData, State> = {
...
+        type="line"
+        options={{
+          plugins: {
+            title: {
+              display: true,
+              text: "Daily counts",
```
mbhrznr, line 85: **nitpick:** personally i'd probably go w/ a `h1` rather than using `options.plugins.title`, mostly for semantic reasons.

routes/stats.tsx
```diff
@@ -44,89 +44,71 @@ export const handler: Handlers<StatsPageData, State> = {
...
+          },
+          interaction: {
+            intersect: false,
+            mode: "index",
+          },
+          scales: {
+            x: {
+              grid: { display: false },
+            },
+            y: {
```
mbhrznr, line 96: **issue (non-blocking):** by removing `steps: { stepSize: 1 }` we might re-introduce some unwanted behavior as seen in https://github.com/denoland/saaskit/issues/276.  

*however* by having a joint chart containing all datasets it'd be less likely to occur in real life.
from a visual standpoint i'd also much prefer the current solution.

locally i was able to reproduce the behavior by simply clicking on one of the labels.  
there you can see that the y-axis is containing floats again:
Bildschirm­foto 2023-07-05 um 08 30 31

mbhrznr (changes requested):
well done! 👏 👏 

as the `useResizeObserver` doesn't do anything anymore due to the lack of `container`, i'd remove that one.
apart from that, lgtm!

routes/stats.tsx
iuioiua: Good point.

routes/stats.tsx
mbhrznr: `ticks: { precision: 0 }` might be a better, more suitable solution though.

routes/stats.tsx
```diff
@@ -35,98 +35,81 @@ export const handler: Handlers<StatsPageData, State> = {
...
+            },
+            scales: {
+              x: {
+                max,
+                grid: { display: false },
+              },
+              y: {
+                beginAtZero: true,
+                grid: { display: false },
+                ticks: { stepSize: 1 },
```
mbhrznr, line 99: **suggestion:** as mentioned in https://github.com/denoland/saaskit/pull/343#discussion_r1252614079, we should be able to replace `{ stepSize: 1 }` with `{ precision: 0 }` to only have integers.

iuioiua: Fixed.

mbhrznr (approved):
really, really well done! 👏 👏 
let's ship it! 🚀

PR #349, opened 2023-07-06
fix: user link on pricing page

brunocorrea23:


iuioiua (approved):
Good catch! Thank you.

PR #350, opened 2023-07-06
feat: adding notification for new votes

brunocorrea23:
new we only have notifications for new comments on a user's post
this PR adds notifications for new upvotes

### DIFF
--- a/routes/api/vote.ts
+++ b/routes/api/vote.ts
@@ -2,7 +2,12 @@
 import type { HandlerContext, Handlers, PageProps } from "$fresh/server.ts";
 import type { State } from "@/routes/_middleware.ts";
 import { createVote, deleteVote, getItem } from "@/utils/db.ts";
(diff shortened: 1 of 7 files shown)


iuioiua (changes requested):
Thanks for this! Can you please make sure you account for the behaviour pointed out in #351?

routes/api/vote.ts
```diff
@@ -34,10 +39,22 @@ async function sharedHandler(
       status = 204;
       await deleteVote(vote);
       break;
-    case "POST":
+    case "POST": {
       status = 201;
       await createVote(vote);
+
+      if (item!.userId != user!.id) {
```
iuioiua, line 46:
```suggestion
      if (item!.userId !== user!.id) {
```
Hm, I would've thought that checking for `item === null || user === null` above would've negated the need for the need for the non-null assertion operator `!`.

iuioiua (approved):
LGTM! Thanks you.

PR #353, opened 2023-07-06
fix: correct description in README

boscohyun:
SSIA


iuioiua (approved):
LGTM! Good catch. Thank you.

PR #355, opened 2023-07-07
fix(stats): replace chartjs esm.sh imports with jsdelivr

mbhrznr:
closes #354

PR #357, opened 2023-07-07
fix: target `es2022` for `Chart.js` import and upgrade to v4.3.0

### LINKED ISSUE #354: bug: charts appear small
We're getting this even when not on the latest version of Chart.js. CC @mbhrznr 

hunt deno land_stats

iuioiua:
I just realised that this might be the cause of our woes. Previously, the target would've been `deno` as islands are built server-side. Checkout "ESBuild Options" in https://esm.sh/#docs 

Closes #354
Supercedes #355


iuioiua:
CC @mbhrznr. This would be worth integrating into your `fresh_charts` PR.

iuioiua:
It's fixed!
image

PR #358, opened 2023-07-10
feat: rating algorithm

### LINKED ISSUE #234: proposal: allow for a custom ranking algorithm, not just votes
Today DenoHunt uses only upvotes to rank the items. It would be good to be able to support a ranking algorithm that involves the 'age' and 'engagement' of the item/post.

It would be nice to discuss how to do the pagination in this case.

roberto-morado:
Improves item sorting accuracy and relevance based on likes, comments, and recency.

- Replaced `compareScore` with `sortByRating` function to sort items based on rating.
- Utilized ratingAlgorithm.ts for rating calculation.
- Unit tests updated accordingly.

Resolves #234

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -1,5 +1,6 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import { chunk } from "std/collections/chunk.ts";
+import { calculateRating } from "@/utils/ratingAlgorithm.ts";
 
 const KV_PATH_KEY = "KV_PATH";
 let path = undefined;
@@ -561,8 +562,30 @@ export async function getAreVotedBySessionId(
   return items.map((item) => votedItemIds.includes(item.id));
 }
 
-export function compareScore(a: Item, b: Item) {
-  return Number(b.score) - Number(a.score);
+export async function sortByRating(items: Item[]): Promise<Item[]> {
+  const values = items.map((
+    value,
+    index,
+  ) => ({
+    index,
+    item: value,
+    score: 0,
+  }));
+
+  for (const value of values) {
+    value.score = calculateRating(
+      value.item.score,
+      (await getCommentsByItem(value.item.id)).length,
+      value.item.createdAt,
+    );
+  }
+
+  values.sort((a, b) => Number(b.score) - Number(a.score));
+
+  const sortedIndices = values.map((value) => value.index);
+  const sortedArray = sortedIndices.map((index) => items[index]);
+
+  return sortedArray;
 }
 
 // Analytics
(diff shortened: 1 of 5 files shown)


utils/ratingAlgorithm.ts
```diff
@@ -0,0 +1,47 @@
...
+  const likesWeight = 0.1;
+  const commentsWeight = 0.5;
+  const recencyWeight = 0.9;
+
+  const likesScore = likes * likesWeight;
+  const commentsScore = comments * commentsWeight;
+
+  // https://stackoverflow.com/questions/2627473/how-to-calculate-the-number-of-days-between-two-dates
+  const start = getTimeFromDate(new Date());
+  const end = getTimeFromDate(postDate);
```
iuioiua, line 40:
```suggestion
  const start = Date.now();
  const end = postDate.getTime();
```
It appears `getTimeFromDate()` isn't needed.

utils/ratingAlgorithm.ts
iuioiua, line 1: This code can be moved into `db.ts`.

utils/ratingAlgorithm.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+/**
+ * Calculates the rating score for a given post based on user engagement, time spent, and recency.
+ * @param {number} likes - Number of likes for the post.
+ * @param {number} comments - Number of comments on the post.
+ * @param {Date} postDate - Date of the post.
+ * @returns {number} - The rating score of the post.
+ */
+export function calculateRating(
```
iuioiua, line 26:
```suggestion
export function calcRating(
```

utils/db.ts
```diff
@@ -561,8 +562,30 @@ export async function getAreVotedBySessionId(
...
+  ) => ({
+    index,
+    item: value,
+    score: 0,
+  }));
+
+  for (const value of values) {
+    value.score = calculateRating(
+      value.item.score,
+      (await getCommentsByItem(value.item.id)).length,
```
iuioiua, line 578: This is a costly operation.

iuioiua (changes requested):
This functionality can be achieved in a far simpler manner:
1. Add a `commentsCount: number` property to the `Item` interface. This property would need an increment every time a comment is created. This would also require a [migration script](https://github.com/denoland/saaskit/blob/4d652e83825bc269125fca1b15bdb9ec3b2277ba/tools/migrate_kv.ts) to be run.
2. Convert and replace the use of `compareScore()` with `compareRank(item: Item)` to be used as the `compareFn` parameter of [`Array.prototype.toSorted()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted).

iuioiua:
Oh, why the close?

PR #359, opened 2023-07-10
feat: rating algorithm

### LINKED ISSUE #209: proposal: show number of comments of an item on the main page next to the timestamp
Similar to what Hacker News does (https://news.ycombinator.com/)
Screenshot 2023-05-18 at 15 56 20

### LINKED ISSUE #234: proposal: allow for a custom ranking algorithm, not just votes

roberto-morado:
This commit reflects the changes made after reviewing and incorporating feedback.

Resolves #234 

Will also resolve #209


utils/db.ts
```diff
@@ -296,6 +302,45 @@ export function newCommentProps(): Pick<Comment, "id" | "createdAt"> {
...
+  newItemsByTimeRes.commentsCount++;
+  newItemsByUserRes.commentsCount++;
+
+  const com = await kv.atomic()
+    .check(itemRes)
+    .check(itemsByTimeRes)
+    .check(itemsByUserRes)
+    .set(itemKey, newItemRes)
+    .set(itemsByTimeKey, newItemsByTimeRes)
+    .set(itemsByUserKey, newItemsByUserRes)
```
iuioiua, line 339: Can you please ensure all operations in this function happen on the same atomic operation (below)?

utils/db.ts
```diff
@@ -561,8 +606,26 @@ export async function getAreVotedBySessionId(
...
+
+  const likesScore = likes * likesWeight;
+  const commentsScore = comments * commentsWeight;
+  const recencyScore = recencyWeight /
+    (((Date.now() - postDate.getTime()) / DAY) + 1);
+
+  return likesScore + commentsScore + recencyScore;
+}
+
+export function compareRank(a: Item, b: Item): number {
```
iuioiua, line 626:
```suggestion
export function compareRating(a: Item, b: Item): number {
```

utils/db.ts
```diff
@@ -561,8 +606,26 @@ export async function getAreVotedBySessionId(
   return items.map((item) => votedItemIds.includes(item.id));
 }
 
-export function compareScore(a: Item, b: Item) {
-  return Number(b.score) - Number(a.score);
+export function calcRate(
+  likes: number,
+  comments: number,
+  postDate: Date,
+): number {
```
iuioiua, line 613:
```suggestion
export function calcRating(item: Item): number {
```

iuioiua (changes requested):
Looking better. Few tweaks are needed.

roberto-morado:
This will also resolve #209

iuioiua (changes requested):
Looking much better! One thing left. Can you please modify `deleteComment()` to match `createComment()`?

iuioiua:
Oh, I almost forgot - can you please modify `migrateKv()` to update current items with their respective `itemCount`s?

roberto-morado:
> Oh, I almost forgot - can you please modify `migrateKv()` to update current items with their respective `itemCount`s?

What is the purpose of this script?

Is it because items already in the database do not contain the `commentsCount` property?

So I would have to count the number of comments in an item and update the value accordingly?

iuioiua:
Yes and yes 🙂

roberto-morado:
> Oh, I almost forgot - can you please modify `migrateKv()` to update current items with their respective `itemCount`s?

I have a question regarding the migrate_kv.ts file; it would appear that the item ids are hard coded, I'll assume that those are the items that are currently in the prod. database, however it would appear that there exist more that 13 items on the prod. db (website). Would you like me to not touch the ids and only work on updating their key-values?... which in this case I'll be adding the commentsCount property to the item.

iuioiua:
The migration script is used whenever there's a change in the database structure. The script, in its current state, has already been run. You can mould it however you'd like now, including removing those item IDs.

iuioiua:
Did you need a hand with the migration script?

iuioiua:
I've been reconsidering this feature. I'm not entirely convinced that the value it adds, if any, outweighs the added complexity.

Happy to others' thoughts.
CC @mbhrznr, @lino-levan, @brunocorrea23

lino-levan:
I'm pretty neutral on this. I'm not a fan or complexity but this is a nice feature. Up to you and the others.

brunocorrea23:
Yeah.. I agree with you. I'm not so sure about it now.
When I created #234 , I believe we were only sorting by votes, which is bad because very old and irrelevant post might appear first. But now that we have the filters for last week, last month and all time, this becomes less of a problem.

I still think that #209 is a nice feature as it indicates how 'hot' is the topic. If we sliced it only to this issue, would you still think that the value outweighs the complexity?

iuioiua:
Yeah, I like the comment count aspect, but the algorithm functionality may not be needed.

mbhrznr:
as @brunocorrea23 already mentioned, having the filters available reduces the need for the algorithm quite a bit.

however the sorting algorithm itself doesn't seem too complicated, does it?
if we'd slice it to only account for the item's comment count, we'd need the migration and the added complexity in item creation (and deletion) anyway, wouldn't we?

or would we try to find another approach on how to integrate the comment count?

iuioiua:
The implementation of the sorting algorithm is straightforward. It's more the algorithm design that could become tedious. For example, why do the weight coefficients have those values in particular? There isn't a clear objective reason why they would use any given values for this use case (Deno Hunt).

The aim is to have items the user likes more likely to appear at the top. Genuinely, IDK that such a solution is that much better than just using votes.

I can see the comments count as helping with that a bit.

> if we'd slice it to only account for the item's comment count, we'd need the migration and the added complexity in item creation (and deletion) anyway, wouldn't we?

Yep, that's right. But it may be worth it.

iuioiua:
It seems it'll be fine to skip the algorithm functionality on this change, @roberto-morado. However, having the item's comments count is still a good idea.

roberto-morado:
Understood.

Also; I apologize for the late response, I became quite busy at my day job this past week.

Let me know if there is anything else I can help with. I love the app, I'd like to contribute more.

PR #363, opened 2023-07-12
feat: `getToggledStyles()`

### LINKED ISSUE #275: Proposal: highlight current time-ago selector
Today they user has no way to know which is the current time filter being applied to the data.
One way to solve this is to simply highlight one of the time selectors (Last Week, Last Month, All Time).

iuioiua:
This feature adds a way of dynamically styling links according to whether they satisfy a particular condition. The appropriate links have been adjusted to use this new function. This also tweaks the design of the time selector on the homepage.

Before:
image

After  (see "Pricing" link in the top-right corner):
image

Closes #275
PTAL @brunocorrea23


utils/display.ts
```diff
@@ -12,3 +13,22 @@ export function timeAgo(time: number | Date) {
...
+ * @example
+ * ```ts
+ * import { getLinkStyles } from "@/utils/display.ts";
+ *
+ * const activeLinkStyles = getLinkStyles(true);
+ *
+ * const inactiveLinkStyles = getLinkStyles(false);
+ * ```
+ */
+export function getLinkStyles(expr: boolean) {
```
kt3k, line 30: I'd imagine there could be more variations in the future (e.g. disabled, etc). How about making this `getLinkStyles(options: { active: boolean })` to prepare for the future extension?

iuioiua: Are you referring to the [`:disabled` CSS pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled)? If so, this function only takes care of the styles while the element is at its "rest state" (without any of these states active).

iuioiua: Would something like `getStyles(baseStyles: string, toggledStyles: string, expr?: boolean)` do the trick?

iuioiua: I've implemented my above suggestion. WDYT?

iuioiua:
I am merging now, as the function caters to various needs. I will happily accept reviews after the merge.

PR #367, opened 2023-07-13
feat: add styles for visited item links

### LINKED ISSUE #286: Proposal: gray out or hide/disable Hunts that user has already visited
what do you think? that way user who is logged in can filter on Hunts that they haven't seen yet

niklasmtj:
This uses the official color code `#551A8B` defined by the w3. More information and the link to the w3 document can be found in this [StackOverflow answer](https://stackoverflow.com/a/4774037). A tailwindcss color that might come close to it would be `text-purple-900` with `#581C87`.

I just put it on the hunt link `/item` itself and not the external URL of the hunt.

This closes #286.


iuioiua (commented):
I can't see the colour changing on my local machine. Are you able to attach a screenshot of it working?

brunocorrea23:
Don't know if that's the case here, but I was having trouble using the `visited:` tag. It wouldn't work on dark mode, only on light mode (or vice-versa). Don't know why.

niklasmtj:
> Don't know if that's the case here, but I was having trouble using the `visited:` tag. It wouldn't work on dark mode, only on light mode (or vice-versa). Don't know why.

That is actually the same thing that I saw as well. I thought it's only on my machine. It worked for me in dark mode but not in light mode. Will have another look into that.

niklasmtj:
That's actually quite interesting. When using `#551A8B` it does not work as expected. However with `text-[purple]` it does. The hex code works with for me in dark mode but not in light mode. With the current change this should work as expected but please also check it. I don't know if this is an browser issue or something with tailwindcss

iuioiua (changes requested):
This looks great in a light colour scheme. Can you find a colour that works well with a dark colour scheme? We want to maintain a decent colour contrast ratio. Using the same purple colour for light and dark schemes would be cool.

See: https://color.a11y.com/ContrastPair/?bgcolor=101827&fgcolor=800080

niklasmtj:
I am trying to get this to work. I'll check why the hex codes does not work. Because one problem is the options we got when using regular css colors. Those does not really work to use one color for both color schemes. I'll let you know

iuioiua:
To get this landed, we can add a JSDoc comment and create a GitHub issue referencing this colour issue. This must be an upstream issue. Sadly, the [`twind` repo](https://github.com/tw-in-js/twind) repo doesn't appear to be actively maintained right now.

niklasmtj:
Alrighty, I'm actually not quite sure if this is not an a11y thing in those browsers. When using something like a green or yellow color in light mode it actually works. Maybe we have to use two different shades of colors in light / dark mode for now. What do you think?

And I know it's weird, that Firefox got it working but Safari and Chrome simply don't want to show the color in light mode.

iuioiua:
Hi @niklasmtj, check this page out: https://www.w3schools.com/cssref/css_colors.php.

It seems as though `twind` only supports these colours for the `:visited` CSS selector. Let's see if there's a good colour for `visited:dark:text-[<colour>]`.

Note: you may have to run `deno run -A --unstable tools/migrate_kv.ts` to get the site to work locally.

niklasmtj:
This should be fine for now! Lets use `lightpink` for now and keep an eye on the `twind` support of hex colors because from an a11y perspective this is not the best color. I tested a lot of the colors you mentioned two weeks ago and in one of the themes it wasn't perfect but okay to work with.

components/ItemSummary.tsx
```diff
@@ -18,7 +18,7 @@ export default function ItemSummary(props: ItemSummaryProps) {
       <div class="space-y-1">
         <p>
           <a
-            class="hover:underline mr-4"
+            class="visited:text-[lightpink] hover:underline mr-4"
```
iuioiua, line 21:
```suggestion
            class="visited:(text-[purple] dark:text-[lightpink]) hover:underline mr-4"
```
This is what I meant. Sound good?

iuioiua (changes requested):

components/ItemSummary.tsx
niklasmtj: Ahh yes, sure that sounds great! Let's do this!

iuioiua (approved):
LGTM! Thanks!

PR #374, opened 2023-07-14
fix: header breaks on mobile screens

### LINKED ISSUE #345: bug: header breaks on mobile screens
Perhaps, we should do a burger menu, like https://deno.land/. Hopefully, the entire solution can be contained within [`/components/Header.ts`](https://github.com/denoland/saaskit/blob/9294244b55e1214f3a056a3d604a222929fbf71a/components/Header.tsx).
hunt deno land_(iPhone SE)

hashrock:
This pull request introduces a hamburger menu on mobile.

Close #345 

**Implementation notes:**

- I had to use descendant selectors and adjacent selectors to implement a pure CSS hamburger menu. This required using twind-preset-ext in Twind v1.
- Two newly added icons use heroicons.
- Requires few line of inline JS to be accessible with keyboard.

**before:**

image

image

**after:**

image
image
image


components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
+        props.active ? ACTIVE_LINK_STYLES : LINK_STYLES,
+        "relative text-gray-500 px-3 py-4 sm:py-2",
+        props.class,
+      )}
+      aria-label={props.ariaLabel}
+    >
+      {props.name}
+    </a>
+  );
+}
```
iuioiua, line 35: I think it'd be cleaner if we just wrote the menu HTML without iterating over an array of elements. Menu items can vary in requirements, making a flexible `<MenuItem />` challenging to manage past a certain point. We can have a `MENU_ITEM_STYLES` constant for any common styles between these menu items. WDYT?

components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
   LINK_STYLES,
-  NAV_STYLES,
   SITE_BAR_STYLES,
   SITE_NAME,
 } from "@/utils/constants.ts";
 import { stripe } from "@/utils/payments.ts";
-import { Bell, CircleFilled } from "./Icons.tsx";
-import { getToggledStyles } from "@/utils/display.ts";
+import { Bars, Bell, CircleFilled, Cross } from "./Icons.tsx";
+import { cx } from "@twind/core";
```
iuioiua, line 10: Sidenote: this might be able to replace the new `getToggledStyles()`.

components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
+  ariaLabel?: string;
+}
+
+function MenuItem(props: MenuItemProps) {
+  return (
+    <a
+      href={props.href}
+      class={cx(
+        props.active ? ACTIVE_LINK_STYLES : LINK_STYLES,
+        "relative text-gray-500 px-3 py-4 sm:py-2",
```
iuioiua, line 27:
```suggestion
        "relative text-gray-500 py-4 sm:py-2",
```
Perhaps, we don't need any horizontal padding. WYDT?

components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
+          navToggleLabel.addEventListener('keydown', () => {
+            if (event.code === 'Space' || event.code === 'Enter') {
+              navToggleLabel.click();
+              event.preventDefault();
+            }
+          });
+        `}
+      </script>
+      <nav
+        class={"hidden flex-col gap-x-4 divide-y divide-solid sm:(flex items-center flex-row divide-y-0)"}
```
iuioiua, line 112:
```suggestion
        class="hidden flex-col gap-x-4 divide-y divide-solid sm:(flex items-center flex-row divide-y-0)"
```

components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
+        {`
+          const navToggleLabel = document.getElementById('nav-toggle-label');
+          navToggleLabel.addEventListener('keydown', () => {
+            if (event.code === 'Space' || event.code === 'Enter') {
+              navToggleLabel.click();
+              event.preventDefault();
+            }
+          });
+        `}
+      </script>
```
iuioiua, line 110: Would it be better to have the menu as an island? Genuine question.

iuioiua (changes requested):
Looking good! Just a few technical concerns.

brunocorrea23:
Looks great!
Suggestion: as it is such an important element of the UI, would it be better to leave the submit button outside the hamburger menu (ie: always visible)?

components/Header.tsx
```diff
@@ -1,74 +1,130 @@
...
+            {props.hasNotifications && (
+              <CircleFilled class="absolute -top-0.5 -right-0.5 text-primary w-2 h-2" />
+            )}
+          </div>
+        </>
+      ),
+      href: "/account/notifications",
+      ariaLabel: "Notifications",
+    },
+    { name: "Submit a project", href: "/submit", class: "sm:hidden" },
```
iuioiua, line 62:
```suggestion
    { name: "Submit", href: "/submit", class: "sm:hidden" },
```
I'm happy to keep the inner text the same.

iuioiua:
> Looks great! Suggestion: as it is such an important element of the UI, would it be better to leave the submit button outside the hamburger menu (ie: always visible)?

I'd prefer to keep it in the menu as it is now. Instead, perhaps we could keep its button styling to stand out. WDYT?

components/Header.tsx
hashrock: Yes, we can use it. I was just writing code like this at first:

```jsx
class={cx(
  getToggledStyles(
    LINK_STYLES,
    ACTIVE_LINK_STYLES,
    props.url.pathname === "/pricing",
  ),
  NAV_ITEM,
)}
```

I noticed that this code can be rewritten as: 

```jsx
class={cx(
  props.url.pathname === "/pricing" ? ACTIVE_LINK_STYLES : LINK_STYLES
  NAV_ITEM,
)}
```

This cx style description is used in deno deploy and can handle both concatenation and toggle of CSS strings. 
Since this header component uses cx throughout, I removed getToggledStyles as a dependency.

I think it would be better to write getToggledStyles with a ternary operator where cx is used, what do you think?

components/Header.tsx
hashrock: Hmm, maybe removing the padding makes the spacing a little too tight.

image

I think padding is necessary in order to widen the clickable area of the target.

with no padding:

image

with padding:

image

components/Header.tsx
hashrock: I'm a little unsure of this. In [the code I referenced](https://github.com/denoland/doc_components/blob/c46d3cdc9505917f5681805f13111f311779a0c6/header.tsx#L101), we wrote it directly in the onClick handler of the element, but since this caused an error in the latest version of Fresh, I wrote it like that.

If it is set to Island, it will take some time before it can be operated, and it is necessary to consider the time when hydration is not done.

In this case, I just wanted to improve accessibility and didn't plan to add anything here, it seems bit overkill to make it an island, so I put it inline. but I don't have a strong opinion here.

components/Header.tsx
hashrock: Isn't it a little vague what to do if the label is just "Submit"? In PH, the label is only Submit, but the target is selected by hover.

image

Rather, what if the header button to say "Submit a project" when it is wide enough?

hashrock:
image

@iuioiua OK, wdyt?

components/Header.tsx
iuioiua: What about if we got rid of `getToggledStyles()` completely across the codebase and used `cx()` instead?

components/Header.tsx
iuioiua: Fair enough. I didn't feel strongly either.

iuioiua (approved):
This looks amazing 😍

iuioiua:
I'll merge. Let's have the `cx()` discussion elsewhere, it needed.

hashrock:
Thank you!

components/Header.tsx
hashrock: yeah if it's okay, I'll like to help to do this

PR #376, opened 2023-07-18
feat: replace custom chart island with `fresh_charts`

mbhrznr:
now that `fresh_charts` officially supports islands, we can migrate our custom island.


iuioiua (approved):
Awesome! Thank you 🙏🏾

PR #381, opened 2023-07-19
chore: upgrade to Fresh v1.3

iuioiua:


marvinhagemeister (approved):
LGTM, although it's technically not necessary to convert all routes to async routes to update to Fresh 1.3 . Don't mind though 👍

iuioiua:
@marvinhagemeister, PTAL at [the CI failures](https://github.com/denoland/saaskit/actions/runs/5596469928/jobs/10233531742?pr=381). It appears that any async function capable of returning synchronously causes issues for async route components.

iuioiua:
I've had to revert some changes so CI can pass. As previously stated, not all routes had to be converted to async route components. It's not ideal, but not an issue either.

Issue #268, opened 2023-06-19
tracking: improve web performance and practices


iuioiua:
https://saaskit.deno.dev/ should have the best possible score on https://pagespeed.web.dev/

This is partly dependent on #260.

brunocorrea23:
This is the current lighthouse score https://saaskit.deno.dev/.

<img width="776" alt="Screenshot 2023-06-21 at 21 57 44" src="https://github.com/denoland/saaskit/assets/8139585/173998ff-3fea-4d24-abe6-947b7a90fb5b">

<img width="782" alt="Screenshot 2023-06-21 at 21 57 52" src="https://github.com/denoland/saaskit/assets/8139585/8f16468c-226d-47bc-9b1f-2bac782be243">


The only score that is not 100 is the SEO score for mobile. This happens because of the size of tap area for some of the elements. Not sure if there is any way to solve this without changing the design of the page (spacing the elements). Also, not sure if we want to do it. 

<img width="778" alt="Screenshot 2023-06-21 at 21 58 57" src="https://github.com/denoland/saaskit/assets/8139585/8e8a6629-b37f-491e-ab91-4c5c41b25dd7">

iuioiua:
I'm happy to increase these items' margins and possibly font size.

brunocorrea23:
@iuioiua , just checking with you.. are you feeling that the app is slow?
On my local machine it is super fast, but on https://hunt.deno.land/ it is taking longer to respond.

lighthouse doesn't seem to be complaining about it, as the performance score is still high, but:

it is taking 1s for server response.
<img width="785" alt="Screenshot 2023-07-26 at 23 02 44" src="https://github.com/denoland/saaskit/assets/8139585/42029816-8886-4901-b918-2103e6863a3d">

there seems to be opportunities on improving server response time
<img width="782" alt="Screenshot 2023-07-26 at 23 03 33" src="https://github.com/denoland/saaskit/assets/8139585/9ff29d4d-87b8-49fa-971d-7dedb15654ee">

TTFB is also not ideal. neither on mobile nor on desktop.
<img width="769" alt="Screenshot 2023-07-26 at 23 04 12" src="https://github.com/denoland/saaskit/assets/8139585/a6d8ed9e-98a3-4aa4-b545-0d5f6466aa27">
[...]

PR #383, opened 2023-07-20
fix: avatar resolution

brunocorrea23:
This will bring the "Best Practices" score from 92 to 100. #268 

Screenshot 2023-07-19 at 21 01 16


iuioiua (approved):
Excellent! Thank you, Bruno.

PR #385, opened 2023-07-20
feat: `<GitHubAvatarImg />`

iuioiua:
This change adds a new size-adjustable and class-customisable `<GitHubAvatarImg />`. It's dynamically generated using the user's login (AKA username). As a result, the `avatarUrl` property from the `User` interface has been removed.

The migration script updates all user objects without the `avatarUrl` property.

Please review, if you'd like, @brunocorrea23.


brunocorrea23:
Seems to be much cleaner!
Only question is if we would like to support a custom avatar img in the future (a user uploaded one?)? Would it be good to have the avatarUrl connected to the user entity on the db?

iuioiua:
Let's cross that bridge once we come to it 🙂

PR #386, opened 2023-07-20
feat: users can delete own item

### LINKED ISSUE #375: proposal: delete item functionality
This should only be able to be done with ones own item. I'm happy to provide guidance.

brunocorrea23:
Closes #375 

Two questions:
- Is this too much a work around? (this = handling two diff actions on a rounds POST handler. Should we create an island+api route to handle the item deletion? not sure about the best practice here)
- Today the item deletion code in db.ts does not handle the deletion of votes or comments associated with that item. Should it?

### DIFF
--- a/components/Icons.tsx
+++ b/components/Icons.tsx
@@ -151,3 +151,27 @@ export function Cross(props: JSX.SVGAttributes<SVGSVGElement>) {
     </svg>
   );
 }
+
+export function Trash(props: JSX.SVGAttributes<SVGSVGElement>) {
+  return (
+    <svg
+      {...props}
+      xmlns="http://www.w3.org/2000/svg"
+      width="24"
+      height="24"
+      viewBox="0 0 24 24"
+      stroke-width="2"
+      stroke="currentColor"
+      fill="none"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+    >
+      <path stroke="none" d="M0 0h24v24H0z" fill="none"></path>
+      <path d="M4 7l16 0"></path>
+      <path d="M10 11l0 6"></path>
+      <path d="M14 11l0 6"></path>
+      <path d="M5 7l1 12a2 2 0 0 0 2 2h8a2 2 0 0 0 2 -2l1 -12"></path>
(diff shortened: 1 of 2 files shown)


iuioiua:
> * Is this too much a work around? (this = handling two diff actions on a rounds POST handler. Should we create an island+api route to handle the item deletion? not sure about the best practice here)

Yes, this does seem a little hacky. I think it should:
- Be an island
- Only appear for the user that created the item
- Use [`confirm()`](https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm) method to confirm whether the user really wants to delete the user
- Use a new `DELETE /item/:id` route
- Redirect to the homepage once the action is complete

> * Today the item deletion code in db.ts does not handle the deletion of votes or comments associated with that item. Should it?

Yeah, it should. I think we should perhaps have these done in separate `deleteItemVotes()` and `deleteItemComments()` functions as not to overcrowd `deleteItem()`. WDYT?

iuioiua:
Actually, are you able to pause work on this PR? I'm working on a series of PRs that should simplify this implementation.

brunocorrea23:
I'll wait! let me know when you've finished

roberto-morado:
### Make this feature _only_ available to premium 🦕 users

iuioiua:
I agree! 👍🏾

brunocorrea23:
@iuioiua , are you still working on those PRs you've mentioned or can I start working on this PR again?

iuioiua:
Thanks for following up! I still need to complete all the details, but I've opened issue #400. Any help would be appreciated 🙂

roberto-morado:
### I can help, assign as you deem appropriate

iuioiua:
Hi Bruno, actually, on second thought, I'd like to delay having this feature. The reason is that we're aiming to launch v1 soon. And v1 needs to have a certain degree of polish. With every new feature, we add to the list of features that need to be polished. Hence, me previously saying we should pause on new features and instead focus on buttoning things up. I want to close this PR without merging if you are okay with it. If I were you, I'd still keep the branch so that you can revisit it after v1. WDYT?

brunocorrea23:
Makes sense! No worries, @iuioiua !

PR #389, opened 2023-07-21
refactor: use `userLogin` property in `Comment` interface

iuioiua:
This change replaces the `userId` property on the `Comment` interface with `userLogin`. The benefit of this is that we eliminate KV-flavoured joins from DB logic, making it much more straightforward.

This is part of a series of PRs moving towards removing the `User["id"]` property from the codebase. Reviewers, please pay close attention to the migration script, which migrates the old comment interface to the new one.

CC @brunocorrea23, @mbhrznr

### DIFF
--- a/routes/item/[id].tsx
+++ b/routes/item/[id].tsx
@@ -12,7 +12,6 @@ import {
   getAreVotedBySessionId,
   getCommentsByItem,
   getItem,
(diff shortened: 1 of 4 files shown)


kt3k (approved):
LGTM

PR #393, opened 2023-07-24
refactor: rework notifications pages

iuioiua:
This change simplifies the logic around the notifications page and makes it less hacky. A few styling changes have also been made.

### DIFF
--- a/fresh.gen.ts
+++ b/fresh.gen.ts
@@ -20,13 +20,14 @@ import * as $14 from "./routes/dashboard/index.tsx";
 import * as $15 from "./routes/feed.ts";
 import * as $16 from "./routes/index.tsx";
 import * as $17 from "./routes/item/[id].tsx";
-import * as $18 from "./routes/notifications/_middleware.ts";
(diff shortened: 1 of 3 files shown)


iuioiua:
@marvinhagemeister, type checks aren't liking `/routes/notifications/index.tsx`. Is this a bug or error on my part?

kt3k:
There seems something wrong in `AsyncRoute` typing https://github.com/denoland/fresh/blob/9cae8fdd5cb31f920efc10402d01da31ab9e90ac/src/server/types.ts#L138

Specifying `SignedInState` in `ctx: RouteContext<unknown, SignedInState>` makes `NotificationsPage` incompatible with `AsyncRoute` (I guess this should be fixed in fresh side)

iuioiua:
Nice catch! Fix incoming.

iuioiua:
Hm... I remove the state type and still have the same issue...

iuioiua:
I've reverted to the handler method for now. I can refactor once the fix has been published to Fresh.

PR #395, opened 2023-07-24
refactor: redesign submit page

iuioiua:
Hash: I feel like this is missing something, but I don't know what. Any ideas, or do you think it's good?
Andy: I included minimal copy, which looks sufficient to me.

Before:
image

After:
image

### DIFF
--- a/components/Icons.tsx
+++ b/components/Icons.tsx
@@ -95,6 +95,47 @@ export function Check(props: JSX.SVGAttributes<SVGSVGElement>) {
   );
 }
 
+export function CheckCircle(props: JSX.SVGAttributes<SVGSVGElement>) {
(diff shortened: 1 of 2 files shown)


hashrock:
let me do some tweaks. please feel free to revert some changes you don't want.

- Limited content width. If white space is conspicuous, it may look better if max-width is specified.
- Wider margins throughout.
- Added labels to text fields. For accessibility, placeholder attributes should not be used in place of labels.
- It looked a little odd that the text in the button was left aligned, so I centered it.

Preview:

image

hashrock (approved):

iuioiua:
Looks amazing! Thanks @hashrock!

PR #396, opened 2023-07-26
feat: users table

### LINKED ISSUE #370: proposal: add simple users table under charts in admin dashboard
- can we add a users table beneath the charts with some fake users? use case would be the Owner can see/edit/whatever on the SaaS users (see the gif too)

(as example)

image

iuioiua:
This change also:
* Makes the site text size slightly smaller. Previously, it seemed too big.
* Adds a `<TabsBar />` component.

Closes #370

localhost_8000_dashboard_users (2)


hashrock (approved):
LGTM!

iuioiua:
@hashrock, would you happen to have any suggestions regarding my previous comment?

hashrock:
@iuioiua hmm I'm not familiar with table styling, so I'm not sure how to solve it.

just a workaround, but how about combine these columns?

image

iuioiua:
> @iuioiua hmm I'm not familiar with table styling, so I'm not sure how to solve it.
> 
> just a workaround, but how about combine these columns?
> 
> image

Done!

PR #401, opened 2023-07-28
refactor: remove `user.id` use and rework vote logic

### LINKED ISSUE #400: proposal: remove use of `user.id`
I propose we remove all use of `user.id` in favour of `user.login`. This would:
1. Reduce the number of unique identifiers in the `user` object from 2 to 1.
2. Remove the need for KV-flavoured joins.
3. Enable pagination to be implemented in a less hacky way than it currently is.

E.g. if this were done, this would remove the need for [`getManyUsers()` in `/routes/index.tsx`](https://github.com/denoland/saaskit/blob/8476598a8fd5ff3bed41244afe03f518ae342571/routes/index.tsx#L50C30-L50C42).

This would involve refactoring item, vote, user and notification logic. However, the difficult part is migrating existing data cleanly in KV.

Any help with this would be greatly appreciated. Please be sure to let us know if you'd like to help here.

iuioiua:
Vote logic was challenging to deal with, hence the rework. The most involved part of this PR is the migration script. In summary:
* All notifications will be deleted, which, IMO, is ok.
* User entries are untouched.
* Items and votes entries are migrated according to their new respective interfaces.

Reviews, please pay close attention to the migration script. It aims to preserve item, vote and user data. This script has worked multiple times on my local machine, but I still want it checked carefully. To test:
1. On the main branch, run `deno task db:reset` and then `deno task db:seed`.
2. Run `deno task start`, sign in, vote and comment on random items, noting the changes.
3. Switch to this branch and run `deno run -A --unstable tools/migrate_kv.ts`.
4. Run `deno task start` and ensure all data is preserved.

Please review, if willing and able, @brunocorrea23 and @mbhrznr.

Closes #400

### DIFF
--- a/routes/_404.tsx
+++ b/routes/_404.tsx
@@ -1,10 +1,10 @@
-import { HEADING_WITH_MARGIN_STYLES, LINK_STYLES } from "@/utils/constants.ts";
+import { HEADING_STYLES, LINK_STYLES } from "@/utils/constants.ts";
 
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 export default function NotFoundPage() {
   return (
     <main class="flex-1 p-4 flex flex-col justify-center text-center">
-      <h1 class={HEADING_WITH_MARGIN_STYLES}>Page not found</h1>
+      <h1 class={HEADING_STYLES}>Page not found</h1>
       <p>
         <a href="/" class={LINK_STYLES}>Return home ›</a>
       </p>
--- a/routes/_500.tsx
+++ b/routes/_500.tsx
@@ -1,14 +1,14 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import { ErrorPageProps } from "$fresh/server.ts";
-import { HEADING_WITH_MARGIN_STYLES, LINK_STYLES } from "@/utils/constants.ts";
+import { HEADING_STYLES, LINK_STYLES } from "@/utils/constants.ts";
 
 export default function Error500Page(props: ErrorPageProps) {
   return (
-    <main class="flex flex-col justify-center p-4 text-center space-y-4">
-      <h1 class={HEADING_WITH_MARGIN_STYLES}>Server error</h1>
+    <main class="flex flex-1 flex-col justify-center p-4 text-center space-y-4">
+      <h1 class={HEADING_STYLES}>Server error</h1>
       <p>500 internal error: {(props.error as Error).message}</p>
       <p>
-        <a href="/" class={LINK_STYLES}>Return home</a>
+        <a href="/" class={LINK_STYLES}>Return home ›</a>
       </p>
     </main>
   );
--- a/tools/migrate_kv.ts
+++ b/tools/migrate_kv.ts
@@ -41,23 +41,12 @@ async function migrateVote(voterUserId: string, oldItem: OldItem) {
     vote.item.id,
   ];
   const itemsByUserKey = ["items_by_user", vote.item.userLogin, vote.item.id];
-  const [itemRes, itemsByTimeRes, itemsByUserRes] = await kv.getMany([
-    itemKey,
-    itemsByTimeKey,
-    itemsByUserKey,
-  ]);
-  assertIsEntry(itemRes);
-  assertIsEntry(itemsByTimeRes);
-  assertIsEntry(itemsByUserRes);
 
   const votesKey = ["votes", vote.id];
   const votesByItemKey = ["votes_by_item", vote.item.id, vote.id];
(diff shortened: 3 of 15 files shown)


routes/_404.tsx
iuioiua, line 1: Passing-by changes.

routes/_500.tsx
iuioiua, line 1: Passing-by changes.

tools/migrate_kv.ts
```diff
@@ -1,14 +1,140 @@
...
+  const votedItemsByUserIter = kv.list<OldItem>({
+    prefix: ["voted_items_by_user"],
+  });
+  for await (const entry of votedItemsByUserIter) {
+    promises1.push(migrateVote(entry.key[1].toString(), entry.value));
+  }
+  await Promise.all(promises1);
+
+  // Items
+  const promises2 = [];
```
iuioiua, line 129: Both arrays of promises, `promises1` and `promises2`, must resolve sequentially. Vote migrations cover all items that have votes. Item migrations cover all remaining items that have no votes.

Using `Promise.all()` is optimal for performance.

tools/migrate_kv.ts
iuioiua, line 1: Note to self: `voted_items_by_user` and `voted_users_by_item` indexes should be deleted once this PR is merged, as they are no longer used.

tools/migrate_kv.ts
```diff
@@ -1,14 +1,140 @@
...
+
+type OldItem = Omit<Item, "userLogin"> & { userId: string };
+type OldUser = User & { id: string };
+
+async function migrateVote(voterUserId: string, oldItem: OldItem) {
+  const [creatorUserRes, voterUserRes] = await kv.getMany<OldUser[]>([
+    ["users", oldItem.userId],
+    ["users", voterUserId],
+  ]);
+  assertIsEntry(creatorUserRes);
```
iuioiua, line 20: The distinction between these two types of users must remain correct throughout this function.

tools/migrate_kv.ts
```diff
@@ -1,14 +1,140 @@
...
+    .set(itemsByUserKey, vote.item)
+    .set(votesKey, vote)
+    .set(votesByItemKey, vote)
+    .set(votesByUserKey, vote)
+    .commit();
+
+  if (!res.ok) throw new Error(`Failed to set vote: ${vote}`);
+}
+
+async function migrateItem(oldItem: OldItem) {
```
iuioiua, line 72: This is run after `migrateVote()` and only takes care of those items with no votes.

iuioiua (commented):
Some comments to understand the changes. There are other minor changes throughout this PR which are self-explanatory.

mbhrznr:
thank you for providing the comments alongside the changes, which helped a lot in understanding the broader context.  
also thank you for providing the steps to get this going!

i was able to get through the steps w/o any issues.
resetting, seeding and running the migration worked like a charm.
the votes i initially added were also persisted and still existed after i run the migration script.

however i've noticed the following:

1. **downvoting** an **upvoted** item removes the vote from the counter, however on refresh the post is still marked as **upvoted**. this way an item can be downvoted to `0`.

1. **upvoting** an item, which you **already interacted with** increments the counter and updates the color in an optimistic update, yet on refresh the vote is gone. initially i thought this could just be the default browser behavior of re-submitting the form, however the same happens whenever i navigate to the page, e.g. via click on the logo. 

1. trying to **upvote** an existing item, which you **haven't interacted with before**, throws an `AssertionError`, which seems to be coming from the `assertIsEntry` check:
    ```sh
    An error occurred during route handling or page rendering. AssertionError: Expected actual: null not to be: null: KV entry not found
    ```

1. initially upvoting any items which are **created after the migration** works as expected. downvoting that exact same item still shows the item as upvoted when i navigate to the page (see 1.).

i will also do a more thorough review of the code changes but wanted to drop a note about my findings asap.

iuioiua:
I've fixed the migration script by removing what may be unneeded assertions in it. @kt3k and @mbhrznr, can you please re-run the migration script with the new changes? This likely fixes the issues outlined in the previous comment.

kt3k (approved):
Ran migration on prod KV snapshot successfully and updated DB seems working without problem. LGTM

mbhrznr:
@iuioiua thanks for the update!

just tested the functionality from resetting to seeding to voting and migrating.
yet it looks like i can still reproduce all of the findings in the initial comment.
did any of these occur on your end(s) as well?
Bildschirm­foto 2023-08-04 um 16 15 31

iuioiua:
I think the `seed` task may cause the issue you're experiencing. I just tested using the actual production data for the site, and it works fine! I'll run the migration and go from there. Thanks for your help, guys!

PR #402, opened 2023-07-28
feat: `<Share />`

iuioiua:
This component dynamically generates links for sharing the current content on the major social media platforms.


hashrock (approved):
LGTM

PR #406, opened 2023-08-02
fix: use TabsBar style at TimeSelector

hashrock:
I like TabsBar style that we're using on the dashboard, so I replaced with it.
It looks different from Header Nav, and I think the selection state is easy to see. wdyt?

Before: 
image

After: 
image

image


iuioiua (approved):
I love it! I was going to do this myself once I had the chance. Thank you.

PR #407, opened 2023-08-02
doc: add a link to auth setup guide

hashrock:
Just an idea, how about issuing an instruction when `GITHUB_CLIENT_ID` and `GITHUB_CLIENT_SECRET` are not set? I think It reduces the feeling of emptyness after initial setup, and it will prevent users from getting lost in both the local environment and the Deno Deploy environment.

If these environment variables are not required, I think this PR is not appropriate and should be closed.

Feel free to rewrite the text if you like this idea.

image

image

### DIFF
--- a/routes/index.tsx
+++ b/routes/index.tsx
@@ -25,7 +25,7 @@ interface HomePageData extends State {
   areVoted: boolean[];
 }
 
-const needsSetup = Deno.env.get("GITHUB_CLIENT_ID") === undefined ||
+const NEEDS_SETUP = Deno.env.get("GITHUB_CLIENT_ID") === undefined ||
   Deno.env.get("GITHUB_CLIENT_SECRET") === undefined;
 
 function calcTimeAgoFilter(url: URL) {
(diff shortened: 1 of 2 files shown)


routes/index.tsx
```diff
@@ -54,8 +55,17 @@ export const handler: Handlers<HomePageData, State> = {
       ctx.state.sessionId,
     );
     const lastPage = calcLastPage(allItems.length, PAGE_LENGTH);
+    const needsSetup = Deno.env.get("GITHUB_CLIENT_ID") === undefined ||
```
iuioiua, line 58: Perhaps, let's move this inline with the component itself. That way, we don't have to pass it from the handler to the component. WDYT?

iuioiua (commented):
I love this! Looks slick!

routes/index.tsx
hashrock: Oh, I didn't think I could do that. moving into component doesn't work but I just noticed I can do that with declaring it at top level. Please review the code for any problems.

hashrock:
I'm a little worried about seeing the environment variables only at startup, but I think Deploy probably restarted the instance after rewriting the environment variables. If Deno Deploy behaves strangely, please revert the last commit

iuioiua:
Actually, doing this check at startup is a better idea. Nice!

iuioiua (approved):
LGTM!

Issue #73, opened 2023-04-12
proposal: add admin dashboard for SaaS owner that enables CRUD operations on users/orgs db (think PHP my admin)


iuioiua:

brunocorrea23:
@iuioiua , I'd love to see this on v1 as I believe this is a core part for many SaaS products.

But I'm curious to know how you are thinking about the concept of orgs for Deno Hunt. What would users that belong to the same org share?

iuioiua:
Frankly, I'd have to think about how we do this. At the moment, there's plenty of work to get done before v1. I think this can wait till after v1.

brunocorrea23:
ah, got it! sorry, I thought that the v1 label you've added meant that you were prioritizing this task to get into that version.
what do you mean by this v1 label?

iuioiua:
Ah, I must've accidentally added the v1 label. I've removed it now. I think having orgs and an admin dashboard with full read/write functionality is not needed for v1.

PR #409, opened 2023-08-02
feat: admin pannel

deer:
closes https://github.com/denoland/saaskit/issues/73

Hey @iuioiua, this is incredibly rough right now, but you encouraged me to open something so we can iterate on the idea.

With these changes, then I can go to the following routes:
* http://localhost:8000/models
* http://localhost:8000/items
* http://localhost:8000/items/new

There are currently two major issues:
* The styling is horrible. I need to get the plugin to accept styles that get applied to the routes it creates.
* The plugin only works with pentagon right now. So most of the functionality doesn't work! I'm at least glad that I can show the items, otherwise I wouldn't even open this. Ideally the user of the plugin could map their own CRUD functions to types, instead of forcing the plugin user to create (possibly unnecessary/unwanted) pentagon models.

I will get to these enhancements at some point and then post an update here. Hopefully that will get rid of the pentagon and zod imports.

After those are dealt with, I can turn to a better layout (likely some sort of table) and then security / permissions.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -18,7 +18,7 @@
   },
   "imports": {
     "@/": "./",
-    "$fresh/": "https://deno.land/x/fresh@1.3.0/",
+    "$fresh/": "https://raw.githubusercontent.com/deer/fresh/feat_plugin_islands/",
     "$gfm": "https://deno.land/x/gfm@0.2.4/mod.ts",
     "preact": "https://esm.sh/preact@10.15.1",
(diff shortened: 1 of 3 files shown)


iuioiua:
Interested to see what this evolves into!

iuioiua:
Hey Reed, are you cool if we close this PR without merging for now? We can revisit once v1 has landed.

deer:
Not at all a problem. Work picked up and I unfortunately had less time to contribute to this than I had hoped.

Issue #346, opened 2023-07-05
tracking: design polish


iuioiua:
It would be good to get a 2nd opinion on the design of some of these pages. Any suggestions would be welcome.
- [ ] [/](https://hunt.deno.land/) (item summary spacing can probably be improved)
- [ ] [/stats](https://hunt.deno.land/stats)
- [ ] [/blog](https://hunt.deno.land/blog)
- [ ] [/account](https://hunt.deno.land/account)
- [x] [/submit](https://hunt.deno.land/submit) (feels a bit empty)
- [x] [/user/:login](https://hunt.deno.land/user/iuioiua) (perhaps, the profile details could be more prominent)
- [x] [/pricing](https://hunt.deno.land/pricing)

Todo:
- #467
- #468
- #469

brunocorrea23:
Adding to this here just two things that might make sense:
- #268 the last thing that lighthouse is complaining in order to get a perfect score is adjusting the size of the tap areas on mobile
- The links on the new pricing page are hard to see on dark mode:
<img width="338" alt="Screenshot 2023-07-05 at 20 39 11" src="https://github.com/denoland/saaskit/assets/8139585/6e4d3ec6-ebd6-4b5e-9e1b-8c23b8d5ba4b">

hashrock:
These are just a suggestion, not a strong opinion.

# Overall / nav

- A little white space stands out, so it might be better to use the space loosely
- Add a little more space between the header and the main content.

# Root

- increase the spacing between each item
- Increase title font size for each item
- Added selected state to Last Week / Last Month / All time buttons (logically tab / radio buttons)
- Add buttons to pagination UI (I usually refer [tailwind UI](https://tailwindui.com/components/application-ui/navigation/pagination))
- upvote button should be implemented with button element (for accessibility, keyboard operation)

# /stats

- Maybe create a section to display numbers such as today's total visitors, weekly total visitors, etc

# /blog

- Increase font size of main content

# /item/:itemId

- Increase font size of title text etc
- maybe add a back link (you can go back by clicking the logo, but to clarify the relationship between collection view and single view)

# submit

feels a bit empty, but I can't think of a way to improve it since there's nothing to display.

hashrock:
I'm not a big fan of this, but it can be used as an accent.

<img width="1094" alt="image" src="https://github.com/denoland/saaskit/assets/3132889/77c833e1-27b2-4aee-a11f-e02189aa0415">


[...]

PR #410, opened 2023-08-05
feat: improve page selector UI

brunocorrea23:
Changing Page Select (pagination) UI as suggested by @hashrock in #346 

Before:
Screenshot 2023-08-05 at 11 45 12

Now:
Screenshot 2023-08-05 at 11 45 27


hashrock (approved):
LGTM!

brunocorrea23:
@iuioiua , what do you think about this proposal here?

iuioiua:
Hi Bruno, sorry for the delay in getting back to you. Check out #414. There I lay out my reason as to why we should go back to `cursor`-based pagination. While I love the look of what you designed, it conflicts with my proposal.

brunocorrea23:
Makes sense! we should use the cursor for pagination. 
But I'd still suggest we use the numbers pagination. We would have to adapt it so that it doesn't load all the info at once, but we can make it work.
Let me take a look at it and I'll send a PR.

PR #411, opened 2023-08-07
refactor: use async route components

iuioiua:
I want to perform this same refactor across the codebase as it's much cleaner. Any reason why this doesn't work, @deer, @kt3k, @marvinhagemeister? I thought we fixed this in https://github.com/denoland/fresh/pull/1510.

Note: the tree used here is the latest as of this PR's creation.


deer:
I spent some time looking at this and also didn't make any progress. All I could do was reproduce it with a simpler case:
```tsx
import type { RouteContext } from "$fresh/server.ts";
import { AsyncRoute } from "$fresh/src/server/types.ts";

type SimpleState = {
  asdf: boolean;
};

const ManagePage: AsyncRoute = async function (
  _req: Request,
  _ctx: RouteContext<any, SimpleState>,
) {
  return new Response();
};

export default ManagePage;
```
and then we get:
```
Type '(_req: Request, _ctx: RouteContext<any, SimpleState>) => Promise<Response>' is not assignable to type 'AsyncRoute'.
  Types of parameters '_ctx' and 'ctx' are incompatible.
    Type 'RouteContext<any, Record<string, unknown>>' is not assignable to type 'RouteContext<any, SimpleState>'.
      Type 'RouteContext<any, Record<string, unknown>>' is not assignable to type 'Omit<HandlerContext<any, SimpleState>, "render">'.
        Types of property 'state' are incompatible.
          Property 'asdf' is missing in type 'Record<string, unknown>' but required in type 'SimpleState'.deno-ts(2322)
```
(At least I think this is the same issue and a simpler case.) Time to do something else all day and reconsider this tonight.

iuioiua:
Yeah, this doesn't work either:
```ts
const ManagePage: AsyncRoute<any, { asdf: boolean }> = async (
  _req: Request,
  _ctx: RouteContext,
) => {
  return new Response();
};

export default ManagePage;
```

deer:
(insert something intelligent about contravariant parameter types which solves the problem here) 🤷‍♂️

iuioiua:
I am closing this for now. We can revisit this once the issue is fixed.

marvinhagemeister:
It's on my to-do list to fix the type issue this week

Issue #414, opened 2023-08-10
proposal: cursor-based pagination


iuioiua:
Currently, pagination uses page numbers, which results in a slightly hacky solution for pagination, retrieving all results on all pages. The codebase should be an example of how KV should be used right. We should instead revert to using the `cursor` parameter from `kv.list()`. This would essentially reduce pagination instances across the site to "Previous" and "Next" links.

Related:
* https://github.com/denoland/saaskit/issues/398#issuecomment-1652903172
* #233
* #242

iuioiua:
Ok, there are two solutions: a "load more" button and an infinite scroll. Infinite scroll is cooler, but a load more button is simpler. Which should we go for? My vote is the load more button.

iuioiua:
We'll go ahead with the load more button.

PR #425, opened 2023-08-15
refactor: cursor-based pagination for comments

iuioiua:
This change introduces a pagination style that's more natural for Deno KV. The kind of pagination is unconventional but remarkably simple, IMO. Once merged, we'll make this the default pagination method for the site.

Note: the last page is empty. This occurs on the last iteration of the async iterator when the [`limit` option](https://deno.land/api@v1.36.1?s=Deno.KvListOptions&unstable=#prop_limit) is divisible by the total number of results of the async iterator. E.g. `limit = 2` and total results = 10, as in this video. Is this a bug?

https://github.com/denoland/saaskit/assets/29347852/d3e144e2-25db-4e5a-82ba-fa4073677850

Towards #414

### DIFF
--- a/routes/items/[id].tsx
+++ b/routes/items/[id].tsx
@@ -1,25 +1,25 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import type { Handlers, RouteContext } from "$fresh/server.ts";
 import ItemSummary from "@/components/ItemSummary.tsx";
-import PageSelector from "@/components/PageSelector.tsx";
 import { BUTTON_STYLES, INPUT_STYLES } from "@/utils/constants.ts";
-import { calcLastPage, calcPageNum, PAGE_LENGTH } from "@/utils/pagination.ts";
 import {
   type Comment,
   createComment,
   createNotification,
   getAreVotedBySessionId,
-  getCommentsByItem,
   getItem,
   getUserBySession,
+  listCommentsByItem,
   newCommentProps,
   newNotificationProps,
   Notification,
+  valuesFromIter,
 } from "@/utils/db.ts";
 import UserPostedAt from "@/components/UserPostedAt.tsx";
 import { redirect } from "@/utils/redirect.ts";
 import Head from "@/components/Head.tsx";
 import { SignedInState } from "@/utils/middleware.ts";
+import PaginationLink from "@/components/PaginationLink.tsx";
 
 export const handler: Handlers<unknown, SignedInState> = {
   async POST(req, ctx) {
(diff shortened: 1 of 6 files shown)


kt3k:
> This occurs on the last iteration of the async iterator when the [limit option](https://deno.land/api@v1.36.1?s=Deno.KvListOptions&unstable=#prop_limit) is divisible by the total number of results of the async iterator. E.g. limit = 2 and total results = 10, as in this video. Is this a bug?

I think one of the common ways to avoid that behavior (probably a bug) is to always get `(number of visible items) + 1` to detect if the next item exist or not.

iuioiua:
> I think one of the common ways to avoid that behavior (probably a bug) is to always get `(number of visible items) + 1` to detect if the next item exist or not.

Nice suggestion! I've added a workaround and raised an issue. I'd like to hear some thoughts on the feel of this new type of pagination. @brunocorrea23 and @mbhrznr, WDYT?

routes/items/[id].tsx
```diff
@@ -91,23 +88,23 @@ export default async function ItemsItemPage(
...
-  const allComments = await getCommentsByItem(id);
-  const comments = allComments
-    .sort((a, b) => b.createdAt.getTime() - a.createdAt.getTime())
-    .slice((pageNum - 1) * PAGE_LENGTH, pageNum * PAGE_LENGTH);
+  const cursor = ctx.url.searchParams.get("cursor") ?? undefined;
+  const iter = listCommentsByItem(itemId, { cursor });
+  const comments = await valuesFromIter(iter);
+
+  /** @todo https://github.com/denoland/deno/issues/20173 */
+  const nextPageIter = listCommentsByItem(itemId, { cursor: iter.cursor });
```
Comment (losfair) [inline routes/items/[id].tsx:100]: Do not do a second list. Instead, increase the limit by one in the first `listCommentsByItem` call and drop the last entry.

Comment (iuioiua) [reply routes/items/[id].tsx]: Done. Is this solution optimal? WDYT?

mbhrznr:
had the chance to checkout and test the feature.
the usage feels nice and the transitions were quick, yet i was wondering if this is the best approach ux wise.

this way the pagination would be uni-directional and there would be no "easy access" to specific comments.
while it would be easy to show oldest comments first via `reverse`, the "in-between" might be harder to access.
and a user would have to start over to (re-)read any comments.

with the `cursor` search param, the url would be less predictable. 
while negligible, it'd make manual url updates almost impossible (when compared to a more "traditional" pagination w/ `page=<number>`.
however as this approach would still provide sharable urls, i'd consider it insignificant.

as always, great work! 👏 👏

losfair (approved):
LGTM for the pagination logic

iuioiua:
Closing in favour of #429 or #438

PR #429, opened 2023-08-17
feat: cursor-based pagination for comments using infinite scroll

iuioiua:
This is an alternative to #425, which uses a REST API and infinite scroll. I'd be interested to hear people's thoughts.

This method appears to work cleaner with Deno KV.

Towards #414

https://github.com/denoland/saaskit/assets/29347852/a521b84d-0180-497d-8d79-ca3d55d47bd4

### DIFF
--- a/islands/CommentsList.tsx
+++ b/islands/CommentsList.tsx
@@ -24,7 +24,7 @@ function CommentSummary(props: Comment) {
 export default function CommentsList(props: { itemId: string }) {
   const commentsSig = useSignal<Comment[]>([]);
   const cursorSig = useSignal<string>("");
-  const observer = useRef<IntersectionObserver>();
+  const observerRef = useRef<IntersectionObserver>();
   const bottomRef = useRef<HTMLDivElement>(null);
 
   async function loadMoreComments() {
@@ -35,7 +35,7 @@ export default function CommentsList(props: { itemId: string }) {
       cursorSig.value,
     );
 
-    if (cursor === "") observer.current?.disconnect();
+    if (cursor === "") observerRef.current?.disconnect();
 
     commentsSig.value = [...commentsSig.value, ...comments];
     cursorSig.value = cursor;
@@ -50,20 +50,18 @@ export default function CommentsList(props: { itemId: string }) {
   }, []);
 
   useEffect(() => {
-    if (!observer.current) {
-      observer.current = new IntersectionObserver(([entry]) => {
+    if (!observerRef.current) {
+      observerRef.current = new IntersectionObserver(async ([entry]) => {
         if (entry.isIntersecting) {
-          loadMoreComments();
+          await loadMoreComments();
         }
       });
     }
 
-    if (cursorSig.value) {
-      observer.current.observe(bottomRef.current!);
-    }
+    observerRef.current.observe(bottomRef.current!);
 
     return () => {
-      observer.current?.disconnect();
+      observerRef.current?.disconnect();
     };
   }, [cursorSig.value]);
 
--- a/islands/CommentsList.tsx
+++ b/islands/CommentsList.tsx
@@ -7,6 +7,7 @@ import CommentSummary from "@/components/CommentSummary.tsx";
 async function fetchComments(itemId: string, cursor: string) {
   let url = `/api/items/${itemId}/comments`;
   if (cursor !== "") url += "?cursor=" + cursor;
+
   const resp = await fetch(url);
   if (!resp.ok) throw new Error(`Request failed: GET ${url}`);
   return await resp.json() as { comments: Comment[]; cursor: string };
@@ -35,7 +36,9 @@ export default function CommentsList(props: { itemId: string }) {
 
   return (
     <div>
-      {commentsSig.value.map((comment) => <CommentSummary {...comment} />)}
+      {commentsSig.value.map((comment) => (
+        <CommentSummary key={comment.id} {...comment} />
+      ))}
       <div ref={bottomRef}></div>
     </div>
   );
--- a/islands/CommentsList.tsx
+++ b/islands/CommentsList.tsx
@@ -1,71 +1,79 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
 import { useSignal } from "@preact/signals";
-import { useCallback, useEffect, useRef } from "preact/hooks";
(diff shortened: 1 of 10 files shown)


mbhrznr:
here's a quick recording.
https://github.com/denoland/saaskit/assets/8133892/2ed272b8-ebe9-4025-aa81-34139a78eb78

islands/CommentsList.tsx
```diff
@@ -0,0 +1,41 @@
...
+async function getComments(itemId: string, cursor: string) {
+  let url = `/api/items/${itemId}/comments`;
+  if (cursor !== "") url += "?cursor=" + cursor;
+  const resp = await fetch(url);
+  if (!resp.ok) throw new Error(`Request failed: GET ${url}`);
+  return await resp.json() as { comments: Comment[]; cursor: string };
+}
+
+export default function CommentsList(props: { itemId: string }) {
+  const commentsSig: Signal<Comment[]> = useSignal([]);
```
marvinhagemeister, line 16:
```suggestion
  const commentsSig = useSignal<Comment[]>([]);
```

islands/CommentsList.tsx
```diff
@@ -0,0 +1,41 @@
...
+    });
+
+    observer.observe(document.querySelector("#bottom")!);
+    return () => observer.disconnect();
+  }, []);
+
+  return (
+    <div>
+      {commentsSig.value.map((comment) => <CommentSummary {...comment} />)}
+      <div id="bottom"></div>
```
marvinhagemeister, line 38: Let's use a `ref` here instead of an id and pass that to the IntersectionObserver. That way this component can be reused multiple times on the same page and we don't need to come up with new names.

islands/CommentsList.tsx
```diff
@@ -0,0 +1,41 @@
...
+      }
+    });
+
+    observer.observe(document.querySelector("#bottom")!);
+    return () => observer.disconnect();
+  }, []);
+
+  return (
+    <div>
+      {commentsSig.value.map((comment) => <CommentSummary {...comment} />)}
```
marvinhagemeister, line 37: Missing `key` prop on `CommentSummary`.

```suggestion
      {commentsSig.value.map((comment) => <CommentSummary key={comment.id} {...comment} />)}
```

marvinhagemeister (commented):
This looks great! Did a quick review, will do one more in depth one tomorrow.

islands/CommentsList.tsx
iuioiua: `<CommentSummary />` doesn't have a `key` property. Am I missing something?

marvinhagemeister: Every vnode has a `key` property. It's not forwarded to the component but instead a hint to the reconciler which element matches to what. It's required for rendering lists where the order of the elements can change. Without it Preact (or any other virtual-dom framework) won't be able to detect if elements have been swapped and will instead destroy and re-create them from scratch.

iuioiua: I see! Added.

iuioiua: 1. Is this practice only good for client-side rendered components?
2. Perhaps, this should be a linting rule.

iuioiua:
> while testing i noticed that the infinite scroll isn't working as intended when `#bottom` is already shown on screen.
> on load it'll show the selected limit of comments, but it won't load any further comments even though it should.

Hm... I'm not sure how to fix this. To reproduce, one can set the `limit` in `/routes/api/items/[id]/comments.ts` to 1. This way, the bottom div is immediately within the viewport upon initial page load.

iuioiua:
Alternatively, we can have a "Load more" button instead of using the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).

mbhrznr:
> Hm... I'm not sure how to fix this. To reproduce, one can set the limit in /routes/api/items/[id]/comments.ts to 1. This way, the bottom div is immediately within the viewport upon initial page load.

refactored the logic a bit and tried the following, which seemed to work quite nicely.
it's making the approach more complicated, but it would fix the issue.

```tsx
import { useSignal } from "@preact/signals";
import { useCallback, useEffect, useRef } from "preact/hooks";
import { Ref } from "preact";
import { Comment } from "@/utils/db.ts";
import UserPostedAt from "@/components/UserPostedAt.tsx";

async function fetchComments(itemId: string, cursor: string) {
  let url = `/api/items/${itemId}/comments`;
  if (cursor !== "") url += "?cursor=" + cursor;

  const resp = await fetch(url);
  if (!resp.ok) throw new Error(`Request failed: GET ${url}`);
  return await resp.json() as { comments: Comment[]; cursor: string };
}

export default function CommentsList(props: { itemId: string }) {
  const commentsSig = useSignal<Comment[]>([]);
  const cursorSig = useSignal("");
  const observer = useRef<IntersectionObserver>();
  const lastElementRef = useCallback((node: HTMLDivElement) => {
    if (!cursorSig.value) return;
    if (observer.current) observer.current.disconnect();

    observer.current = new IntersectionObserver(async ([entry]) => {
      if (entry.isIntersecting) {
        const { comments, cursor } = await fetchComments(
          props.itemId,
          cursorSig.value,
        );

        if (cursor === "") observer.current?.unobserve(entry.target);

        commentsSig.value = [...commentsSig.value, ...comments];
        cursorSig.value = cursor;
      }
    });

    if (node) observer.current.observe(node);
  }, [cursorSig.value]);

  useEffect(() => {
    fetchComments(props.itemId, cursorSig.value).then(
      ({ comments, cursor }) => {
        commentsSig.value = comments;
        cursorSig.value = cursor;
      },
    );
  }, []);

  return (
    <div>
      {commentsSig.value.map((comment) => {
        const isLast =
          comment.id === commentsSig.value[commentsSig.value.length - 1].id;
        const props = {
          class: "py-4",
          key: comment.id,
          ref: isLast ? lastElementRef as Ref<HTMLDivElement> : null,
        };

        return (
          <div {...props}>
            <UserPostedAt {...comment} />
            <p>{comment.text}</p>
          </div>
        );
      })}
    </div>
  );
}
```

iuioiua:
Thanks, @mbhrznr! I've simplified your fix.

iuioiua:
Closing in favour of the simpler "load more" button.

PR #430, opened 2023-08-17
refactor: use `x/tabler_icons_tsx` import

JacobDanton:
Easier to maintain than manually copy-pasted SVGs from various sources.


iuioiua (approved):
Thanks for this! I went further and deleted `/components/Icons.tsx` and imported the components directly from the `x/tabler_icons_tsx` module. Now, LGTM!

Issue #427, opened 2023-08-15
Vote count not working as expected


bminde:
**Describe the bug** A clear and concise description of what the bug is.
I'm able to repeatedly down vote an item.

**To Reproduce** Steps to reproduce the behavior:

1. Go to an item, like https://hunt.deno.land/items/7af60566-c2dc-480a-a77f-e432cb994930
2. Click on vote arrow
3. Reload page
4. Repeat 2 and 3

**Expected behavior** A clear and concise description of what you expected to happen.
1 vote added, 1 vote subtracted, repeatedly

**Screenshots** If applicable, add screenshots to help explain your problem.
![saaskit](https://github.com/denoland/saaskit/assets/8491/a4566cef-2394-4235-8ca1-11a89910a089)

**Desktop (please complete the following information):**

- OS: macOS Ventura 13.5
- Browser Brave 1.56.20, Safari 16.6

iuioiua:
@huai-jie, is this something you'd be happy to look at?

huai-jie:
Sure 👍

huai-jie:
### _Prior Condition: After the user upvote the item_

Due to the 'failed' delete action ([ refer here ](https://github.com/denoland/saaskit/pull/437)) the user always is upvoting the item after the `upvote` action 

The upvote button will only undergo `score --` in the `delete` action and causing this kind of issue

kt3k:
There seem 2 bugs in `DELETE /api/vote` endpoint:

- Because we always create vote.id by `newVoteProps()` even when deleting the vote, the existing vote is not deleted because vote.id doesn't match to what we want to delete.
- In `deleteVote` function in `utils/db.ts`, we don't check existing vote keys (`votesKey`, `votesByItemKey`, `votesByUserKey`). The delete calls for these keys are noop, but these should throw to prevent false delete.

iuioiua:
Thanks, @huai-jie and @kt3k. I'm working through a fix at the moment and realise that the vote functionality changes in #401 are likely a mistake that needs reverting 🙁

`vote.id` seems not to be useful. If we revert, we can then do the following in an easier manner:
1. `POST /api/items/[id]/vote` for the current signed-in user
2. `DELETE /api/items/[id]/vote` for the current signed-in user
3. `GET /api/items/[id]/votes` and the corresponding web page
4. `GET /api/users/[login]/votes` and the corresponding web page

What do we think?

iuioiua:
Perhaps, even:
```ts
[...]

PR #437, opened 2023-08-18
draft: fix vote

huai-jie:
fix https://github.com/denoland/saaskit/issues/427

```
export function newVoteProps(): Pick<Vote, "id" | "createdAt"> {
  return {
    id: crypto.randomUUID(),
    createdAt: new Date(),
  };
}
```

```
 const vote = {
    item,
    userLogin: user.login,
    ...newVoteProps(),
  };
  let status;
  switch (req.method) {
    case "DELETE":
      status = 204;
      await deleteVote(vote);
      break;
```

`vote.id` should not get from `newVoteProps` when undergoing the `delete` action

### DIFF
--- a/components/ItemSummary.tsx
+++ b/components/ItemSummary.tsx
@@ -6,6 +6,7 @@ import UserPostedAt from "./UserPostedAt.tsx";
 export interface ItemSummaryProps {
   item: Item;
   isVoted: boolean;
+  votes: any;
 }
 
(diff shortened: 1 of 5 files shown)


iuioiua:
Thank you! Can you please run `deno fmt` and then `deno task ok`?

iuioiua:
I'm closing this PR in favour of #445. Either way, thank you for your help @huai-jie! Feel free to take a look at that PR and tell me what you think.

PR #443, opened 2023-08-22
feat: add URL validation

### LINKED ISSUE #378: proposal: validate urls
I'd suggest to validate links, so you don't end up with something like this:

image

mbhrznr:
closes #378.  

adds some naive url validations to the submit route.  
adds some tests for these naive validations.

these validations check that:
- the urls start w/ `http:` or `https:`
- the urls don't use local ip addresses and/or `localhost`

### DIFF
--- a/utils/url_validation.ts
+++ b/utils/url_validation.ts
@@ -1,4 +1,4 @@
-// Copyright 2018-2023 the Deno authors. All rights reserved. MIT license.
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
 
 export function isValidUrl(string: string): boolean {
   return URL.canParse(string) && string.startsWith("http");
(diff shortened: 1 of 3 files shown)


utils/url_validation.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export function isValidUrl(string: string): boolean {
+  try {
+    const { protocol } = new URL(string);
+    const pattern = /^(https?):/
+
+    return pattern.test(protocol);
+  } catch (_) {
+    return false;
+  }
```
iuioiua, line 11:
```suggestion
  return URL.canParse(string) && string.startsWith("http");
```
I prefer to avoid regex when possible.

iuioiua (approved):
LGTM! Thanks, Michael!

iuioiua:
Note to self: `URL.canParse()` types are missing.

utils/url_validation.ts
mbhrznr: nice, learned something new!

iuioiua: I avoided using `URL.canParse()` because types are missing in the runtime. Fix incoming: https://github.com/denoland/deno/pull/20244

PR #444, opened 2023-08-22
feat: improved HTTP errors for REST API

iuioiua:
This change adds middleware for the REST API that catches HTTP errors, as defined in the Standard Library, and returns non-HTML responses. This will make future REST API development cleaner.

Currently, responses are still responding with HTML. Perhaps, the middleware is set up incorrectly.


marvinhagemeister (approved):

kt3k (approved):
LGTM

iuioiua:
CI errors fixed. Note to self: always run `deno task start` before pushing commits.

PR #447, opened 2023-08-26
feat: disallow duplicate entries for users

### LINKED ISSUE #377: proposal: disallow duplicate entries
I'd suggest to disallow duplicate entries (within a certain timeframe) or rate limit post functionality heavily to make spamming more difficult.

mbhrznr:
closes #377.  

this disallows users from submitting duplicate entries.  
upon submission all user entries are checked for uniqueness on either `title` or `url`.

### DIFF
--- a/routes/submit/index.tsx
+++ b/routes/submit/index.tsx
@@ -22,25 +22,25 @@ export const handler: Handlers<SignedInState, SignedInState> = {
     const url = form.get("url");
 
     if (typeof title !== "string" || typeof url !== "string") {
-      return new Response(null, { status: 400 });
+      return new Response("URL or Title are either invalid or missing.", { status: 400 });
     }
 
     try {
       if (!isValidUrl(url) || !isPublicUrl(url)) {
-        return new Response(null, { status: 400 });
+        return new Response("URL must be valid and publicly available.", { status: 400 });
       }
     } catch {
-      return new Response(null, { status: 400 });
+      return new Response("An Error occured while validating the URL.", { status: 500 });
     }
 
     const user = await getUserBySession(ctx.state.sessionId);
 
-    if (!user) return new Response(null, { status: 400 });
+    if (!user) return new Response("You must be logged in to submit a new Item.", { status: 401 });
 
     const items = await getItemsByUser(user.login);
 
     if (items.some((item) => item.url === url || item.title === title)) {
-      return new Response(null, { status: 400 });
+      return new Response("You already submitted an Item which contains the same Title or URL.", { status: 409 });
     }
 
     const item: Item = {
--- a/routes/submit/index.tsx
+++ b/routes/submit/index.tsx
@@ -22,27 +22,27 @@ export const handler: Handlers<SignedInState, SignedInState> = {
     const url = form.get("url");
 
     if (typeof title !== "string" || typeof url !== "string") {
-      return new Response("URL or Title are either invalid or missing.", {
+      return new Response("Item URL or title are either invalid or missing.", {
         status: 400,
       });
     }
 
     try {
       if (!isValidUrl(url) || !isPublicUrl(url)) {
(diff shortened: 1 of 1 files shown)


mbhrznr:
`/submit` already has some checks in place, yet we always return an empty response with status `400`.
would it make sense to return more distinct responses in terms of status and (potentially) message?

iuioiua:
Yep! I think a [HTTP 409 status](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/409) with a concise text body message would be fitting 👍🏾

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
     const url = form.get("url");
 
     if (typeof title !== "string" || typeof url !== "string") {
-      return new Response("URL or Title are either invalid or missing.", { status: 400 });
+      return new Response("URL or Title are either invalid or missing.", {
```
iuioiua, line 25:
```suggestion
      return new Response("Item URL or title are either invalid or missing.", {
```

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
...
+        status: 400,
+      });
     }
 
     try {
       if (!isValidUrl(url) || !isPublicUrl(url)) {
-        return new Response("URL must be valid and publicly available.", { status: 400 });
+        return new Response("URL must be valid and publicly available.", {
```
iuioiua, line 32:
```suggestion
        return new Response("Item URL must be valid and publicly available.", {
```

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
...
+          status: 400,
+        });
       }
     } catch {
-      return new Response("An Error occured while validating the URL.", { status: 500 });
+      return new Response("An Error occurred while validating the URL.", {
```
iuioiua, line 37:
```suggestion
      return new Response("An error occurred while validating the URL.", {
```

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
...
+        status: 500,
+      });
     }
 
     const user = await getUserBySession(ctx.state.sessionId);
 
-    if (!user) return new Response("You must be logged in to submit a new Item.", { status: 401 });
+    if (!user) {
```
iuioiua, line 44:
```suggestion
    if (user === null) {
```

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
...
+      return new Response("You must be logged in to submit a new Item.", {
```
iuioiua, line 45:
```suggestion
      return new Response("You must be logged in to submit a new item.", {
```

routes/submit/index.tsx
```diff
@@ -22,25 +22,38 @@ export const handler: Handlers<SignedInState, SignedInState> = {
...
+        status: 401,
+      });
+    }
 
     const items = await getItemsByUser(user.login);
 
     if (items.some((item) => item.url === url || item.title === title)) {
-      return new Response("You already submitted an Item which contains the same Title or URL.", { status: 409 });
+      return new Response(
+        "You already submitted an Item which contains the same Title or URL.",
```
iuioiua, line 54:
```suggestion
        "You already submitted an item which contains the same title or URL.",
```

iuioiua (changes requested):

iuioiua (approved):
I've made some tweaks, and now, LGTM! Thanks, Michael 🙏🏾

Issue #470, opened 2023-09-01
todo: rename and restructure `/tools` folder


iuioiua:
Suggestion:

- [ ] Rename `/tools` to `/tasks`
- [ ] Have each filename match up to the task name. E.g. `db:reset` task for `/tasks/db_reset.ts`.

rj2704:
Hi, I want to work on this issue. Can you please elaborate on the second point that we need to fix?

iuioiua:
Great! Thank you. The `deno.json` file has a number of tasks. Each task correlates with one of the scripts currently in `/tools`. However, their names do not match their file and function names, as given in my example. The task, file and function names matching makes their purposes clearer.

iuioiua:
Hi, @rj2704. Actually, no need to have a main function since each file is a script. Check out #483. It gives an example of what I'm after by having a `db:migrate` task and the `tasks/db_migrate.ts` script.

PR #483, opened 2023-09-02
refactor: use ULID for notifications functionality

### LINKED ISSUE #473: todo: use ULIDs for notification functionality
See #476

iuioiua:
Changes:
1. Notifications use ULIDs for their IDs.
2. Removes the `notifications` index from KV, as it seems unnecessary.
3. Removed the `createdAt` property from notifications.
4. Removes `newNotificationProps()`, as it now seems unnecessary.

Closes #473
Towards #470
CC @lino-levan (this might be a good reference to learn more about ULIDs)


tasks/db_migrate.ts
iuioiua, line 1: Towards #470

utils/monotonic_factory.ts
iuioiua, line 1: `monotonic_factory.ts` is used instead of `ulid.ts` to avoid possible confusion with the new `ulid` import specifier.

deno.json
```diff
@@ -30,7 +31,7 @@
     "@preact/signals-core": "https://esm.sh/*@preact/signals-core@1.2.3",
     "twind-preset-tailwind/": "https://esm.sh/@twind/preset-tailwind@1.1.4/",
     "twind-preset-ext": "https://esm.sh/@twind/preset-ext@1.0.7/",
-    "std/": "https://deno.land/std@0.188.0/",
+    "std/": "https://raw.githubusercontent.com/lino-levan/deno_std/feat-ulid/",
```
iuioiua, line 34: I'll update this once the PR is merged.

iuioiua:
PTAL, @kt3k.

utils/db.ts
```diff
@@ -144,80 +144,74 @@ export function listItemsByTime(options?: Deno.KvListOptions) {
...
     "notifications_by_user",
     notification.userLogin,
-    notification.createdAt.getTime(),
     notification.id,
   ];
 
   const res = await kv.atomic()
-    .delete(notificationsKey)
-    .delete(notificationsByUserKey)
+    .delete(key)
```
kt3k, line 199: Because `.delete` never throws even when the key doesn't exist, I wonder if it's better to `.check` the key as well.

(The bug of vote a while ago was caused by missing check of deleting key of votes)

kt3k:
Left one comment. Other part looks good to me

utils/db.ts
iuioiua: Ah, good point. I've added the check and tweaked tests.

utils/db.ts
```diff
@@ -144,80 +144,75 @@ export function listItemsByTime(options?: Deno.KvListOptions) {
...
+) {
+  const key = [
     "notifications_by_user",
     notification.userLogin,
-    notification.createdAt.getTime(),
     notification.id,
   ];
+  const notificationRes = await kv.get<Notification>(key);
+  if (notificationRes.value === null) {
+    throw new Deno.errors.NotFound("Notification not found");
```
kt3k, line 197: 👍

PR #499, opened 2023-09-05
refactor: remove `getValue()`

### LINKED ISSUE #463: todo: remove `getValue()`
[`getValue()`](https://github.com/denoland/saaskit/blob/0f61d354a36d4f877c58feb162531d3cf087e5f7/utils/db.ts#L16-L22) is too trivial. Best we delete it and just use `entry.value` where needed.

roberto-morado:
fixes #463

### DIFF
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -202,12 +202,11 @@ export async function deleteNotification(
 export async function getNotification(
   notification: Pick<Notification, "id" | "userLogin">,
 ) {
-  const notificationsByIdKey = [
+  const res = await kv.get<Notification>([
     "notifications_by_user",
     notification.userLogin,
     notification.id,
-  ];
-  const res = await kv.get<Notification>(notificationsByIdKey);
+  ]);
   return res.value;
 }
 
@@ -477,26 +476,25 @@ export async function deleteUserBySession(sessionId: string) {
 
 /** @todo Migrate to ["users", login] key */
 export async function getUser(login: string) {
-  const usersByLoginKey = ["users", login];
-  const res = await kv.get<User>(usersByLoginKey);
+  const res = await kv.get<User>(["users", login]);
   return res.value;
 }
 
 export async function getUserBySession(sessionId: string) {
-  const usersBySessionKey = ["users_by_session", sessionId];
-  const eventualRes = await kv.get<User>(usersBySessionKey, {
+  const key = ["users_by_session", sessionId];
+  const eventualRes = await kv.get<User>(key, {
     consistency: "eventual",
   });
-  const res = await kv.get<User>(usersBySessionKey);
-  return eventualRes.value ?? res.value;
+  if (eventualRes.value !== null) return eventualRes.value;
+  const res = await kv.get<User>(key);
+  return res.value;
 }
 
 export async function getUserByStripeCustomer(stripeCustomerId: string) {
-  const usersByStripeCustomerKey = [
+  const res = await kv.get<User>([
     "users_by_stripe_customer",
     stripeCustomerId,
-  ];
-  const res = await kv.get<User>(usersByStripeCustomerKey);
+  ]);
   return res.value;
 }
 
--- a/utils/db.ts
+++ b/utils/db.ts
@@ -198,8 +198,14 @@ export async function deleteNotification(
   if (!res.ok) throw new Error("Failed to delete notification");
 }
 
-export async function getNotification(id: string) {
-  const notificationsByIdKey = ["notifications", id];
+export async function getNotification(
+  notification: Pick<Notification, "id" | "userLogin">,
+) {
+  const notificationsByIdKey = [
+    "notifications_by_user",
+    notification.userLogin,
+    notification.id,
+  ];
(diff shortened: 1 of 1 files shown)


utils/db.ts
```diff
@@ -128,7 +120,8 @@ export async function deleteItem(item: Item) {
 }
 
 export async function getItem(id: string) {
-  return await getValue<Item>(["items", id]);
+  const itemsByIdKey = ["items", id];
+  return (await kv.get<Item>(itemsByIdKey)).value;
```
iuioiua, line 124:
```suggestion
  const res = await kv.get<Item>(["items", id]);
  return res.value;
```
This pattern is used in [the docs](https://deno.land/manual@v1.36.4/runtime/kv). Let's stick to that 🙂

iuioiua (changes requested):
Thanks, @roberto-morado! One main concern.

utils/db.ts
roberto-morado: Gotcha

utils/db.ts
```diff
@@ -128,7 +120,9 @@ export async function deleteItem(item: Item) {
 }
 
 export async function getItem(id: string) {
-  return await getValue<Item>(["items", id]);
+  const itemsByIdKey = ["items", id];
+  const res = await kv.get<Item>(itemsByIdKey);
+  return res.value;
```
iuioiua, line 125:
```suggestion
  const res = await kv.get<Item>(["items", id]);
  return res.value;
```
We can just directly use the key in the argument here.

utils/db.ts
```diff
@@ -208,11 +202,13 @@ export async function deleteNotification(
...
   notification: Pick<Notification, "id" | "userLogin">,
 ) {
-  return await getValue<Notification>([
+  const notificationsByIdKey = [
     "notifications_by_user",
     notification.userLogin,
     notification.id,
-  ]);
+  ];
+  const res = await kv.get<Notification>(notificationsByIdKey);
```
iuioiua, line 210: Ditto. Let's just use the key directly in the argument.

utils/db.ts
```diff
@@ -481,21 +477,27 @@ export async function deleteUserBySession(sessionId: string) {
 
 /** @todo Migrate to ["users", login] key */
 export async function getUser(login: string) {
-  return await getValue<User>(["users", login]);
+  const usersByLoginKey = ["users", login];
+  const res = await kv.get<User>(usersByLoginKey);
```
iuioiua, line 481: Ditto

utils/db.ts
```diff
@@ -481,21 +477,27 @@ export async function deleteUserBySession(sessionId: string) {
...
     consistency: "eventual",
-  }) ?? await getValue<User>(usersBySessionKey);
+  });
+  const res = await kv.get<User>(usersBySessionKey);
+  return eventualRes.value ?? res.value;
 }
 
 export async function getUserByStripeCustomer(stripeCustomerId: string) {
-  return await getValue<User>([
+  const usersByStripeCustomerKey = [
```
iuioiua, line 495: Ditto

utils/db.ts
```diff
@@ -481,21 +477,27 @@ export async function deleteUserBySession(sessionId: string) {
...
 /** @todo Migrate to ["users", login] key */
 export async function getUser(login: string) {
-  return await getValue<User>(["users", login]);
+  const usersByLoginKey = ["users", login];
+  const res = await kv.get<User>(usersByLoginKey);
+  return res.value;
 }
 
 export async function getUserBySession(sessionId: string) {
   const usersBySessionKey = ["users_by_session", sessionId];
```
iuioiua, line 486:
```suggestion
  const key = ["users_by_session", sessionId];
```

utils/db.ts
```diff
@@ -481,21 +477,27 @@ export async function deleteUserBySession(sessionId: string) {
...
-  return await getValue<User>(usersBySessionKey, {
+  const eventualRes = await kv.get<User>(usersBySessionKey, {
     consistency: "eventual",
-  }) ?? await getValue<User>(usersBySessionKey);
+  });
```
iuioiua, line 489:
```suggestion
  });
  if (eventualRes.value !== null) return eventualRes.value;
```

utils/db.ts
```diff
@@ -481,21 +477,27 @@ export async function deleteUserBySession(sessionId: string) {
 
 /** @todo Migrate to ["users", login] key */
 export async function getUser(login: string) {
-  return await getValue<User>(["users", login]);
+  const usersByLoginKey = ["users", login];
+  const res = await kv.get<User>(usersByLoginKey);
+  return res.value;
 }
 
 export async function getUserBySession(sessionId: string) {
```
iuioiua, line 485: Previously, this function performed 1 eventual read and _possibly_ 1 consistent read. Now, with your changes, it performs both every time. Let's fix that up.

iuioiua (changes requested):

iuioiua (approved):
That's it! LGTM. Thank you, @roberto-morado.

PR #500, opened 2023-09-05
Improve user page header

### LINKED ISSUE #468: todo: improve user page header
The current user page header looks like this:
Screenshot 2023-09-01 at 12 05 56 pm

I think it'd look nice if:
- Everything was horizontally aligned to the centre - avatar image on the top, username and badge in the middle and GitHub link on the bottom
- Avatar image was bigger

WDYT, @hashrock?

roberto-morado:
fixes #468 

CC @iuioiua

### DIFF
--- a/utils/http.ts
+++ b/utils/http.ts
@@ -1,14 +1,25 @@
 // Copyright 2023 the Deno authors. All rights reserved. MIT license.
-import type { RedirectStatus, Status } from "std/http/http_status.ts";
+import { RedirectStatus, Status } from "std/http/http_status.ts";
 
 /**
+ * Returns a response that redirects the client to the given location (URL).
+ *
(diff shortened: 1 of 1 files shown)


iuioiua (changes requested):
I think this whole component could be greatly simplified using `w-full`, `text-center` and `space-y-2` classes and a couple of `<p>` elements. Flex is likely not needed. WDYT?

roberto-morado:
What do you think?

iuioiua (changes requested):
Suggestions:
1. Make username text larger, maybe `text-xl` and remove `<strong />`
2. Maybe put username and GitHub logo on the same line

I'm not certain about these. WDYT?

roberto-morado:
Let me create a few alternatives from which you can choose and we work from there 👍

PR #508, opened 2023-09-05
chore: flatten e2e tests

### LINKED ISSUE #496: todo: flatten e2e tests
Right now, we have [/e2e_test.ts](https://github.com/denoland/saaskit/blob/400c29f1482c6ce96571dc4ab47f4a5754a6d28d/e2e_test.ts) which has a single top-level test and then sub-tests. This is inconsistent with the rest of the codebase. There should be one top-level test per route.

mbhrznr:
closes #496


iuioiua (approved):
Very nice! Thank you, @mbhrznr.

Issue #513, opened 2023-09-05
todo(perf): only load `<VoteButton />` island if user is signed in


iuioiua:
Spec:
- Replace with a component identical in look to `<VoteButton />` when the user is not signed in.
- Upon hover, state "Sign in to vote", possibly using the [`title` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title).
- Do not change the mouse state upon hover to avoid misleading the user.

ryanponce:
I took a stab at this issue here. https://github.com/denoland/saaskit/pull/517

iuioiua:
Completed in #517.

PR #517, opened 2023-09-06
perf: `<StaticVoteButton />`

ryanponce:
I created a component called `StaticVoteButton`. Struggled on the same but I'm open to suggestions to make it more clear if needed. The `title` attribute seemed to work well on hover (see attached screenshot).

Screenshot 2023-09-05 at 9 38 36 PM

Related to #513


components/ItemSummary.tsx
```diff
@@ -2,19 +2,25 @@
 import VoteButton from "@/islands/VoteButton.tsx";
 import type { Item } from "@/utils/db.ts";
 import UserPostedAt from "./UserPostedAt.tsx";
+import StaticVoteButton from "./StaticVoteButton.tsx";
 
 export interface ItemSummaryProps {
   item: Item;
   isVoted: boolean;
+  isSignedIn: boolean;
```
iuioiua, line 10: Nice! I wonder if having `isVoted?: boolean` might be simpler, where it's `undefined` if the user is not signed in. WDYT?

routes/users/[login].tsx
```diff
@@ -44,7 +44,8 @@ export default async function UsersUserPage(
 ) {
   const { login } = ctx.params;
   const user = await getUser(login);
-  if (user === null) return await ctx.renderNotFound();
+  const isSignedIn = user !== null;
+  if (!isSignedIn) return await ctx.renderNotFound();
```
Comment (iuioiua) [inline routes/users/[login].tsx:48]:
```suggestion
  if (user === null) return await ctx.renderNotFound();
  const isSignedIn = ctx.state.sessionUser !== undefined;
```

iuioiua (changes requested):
Cleanly done so far! That is a good starting point.

components/ItemSummary.tsx
ryanponce: Yeah, I like that suggestion if reusing the `isVoted` prop. What do you think about `isVoted: boolean | undefined` instead since we probably want that prop to be required?

iuioiua: `isVoted?: boolean` is the same as `isVoted: boolean | undefined` ([reference](https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters)). Having `isVoted?: boolean` would mean it is optional. Btw, if, for whatever reason, it's not simpler, let's stick to using `isSignedIn: boolean`.

iuioiua (approved):
Actually, let's stick with this as #534 will also be using the `isSignedIn` property. Thanks for this!

PR #519, opened 2023-09-06
chore: simplify `tasks/db_seed.ts`

### LINKED ISSUE #471: todo: simplify `/tasks/db_seed.ts`
[This file](https://github.com/denoland/saaskit/blob/b290ffa6cd7273eebbfd63ebba755d572441de0c/tasks/db_seed.ts) appears to be more complicated than it needs to be. My suggestions:
* Remove `batchify()`. Perhaps, instead, use an approach like [this](https://github.com/denoland/saaskit/blob/a1386e93edfca8306bbe8b90f00f1f29c8e0a620/tools/reset_kv.ts#L6-L8) for promises.
* Reduce nesting
* Throw errors instead of `console.error()`ing and returning like what's been done [here](https://github.com/denoland/saaskit/blob/a1386e93edfca8306bbe8b90f00f1f29c8e0a620/tools/seed_submissions.ts#L33-L37).
* Ensure every function is used and needed.

mbhrznr:
closes #471


mbhrznr:
@iuioiua, hope that's what you had in mind.

iuioiua (approved):
LGTM! Thank you.

PR #534, opened 2023-09-07
fix: only call `/api/me/votes` when being signed in

### LINKED ISSUE #522: todo(perf): avoid calling `GET /api/me/votes` if user is not signed in
This call is unnecessary in this case.

mbhrznr:
closes #522.


islands/ItemsList.tsx
```diff
@@ -61,12 +63,17 @@ export default function ItemsList(props: { endpoint: string }) {
...
+      return;
+    }
+
     fetchVotedItems()
       .then((votedItems) =>
         votedItemsIdsSig.value = votedItems.map(({ id }) => id)
       )
       .finally(() => loadMoreItems());
-  }, []);
+  }, [props.isSignedIn]);
```
iuioiua, line 76: Is this needed? I'm not sure it is.

iuioiua: I removed this as I think it's unnecessary. If you think it's necessary, please let me know.

iuioiua (approved):
I've made some edits and now, LGTM! Thank you, Michael.

islands/ItemsList.tsx
mbhrznr: shouldn't be needed, thanks for removing!

PR #536, opened 2023-09-08
fix: `db:migrate` task

iuioiua:
This change aims to eliminate concurrency in the migration script. Excessive concurrent HTTP requests break the migration process (see https://github.com/seanmonstar/reqwest/issues/976).


kt3k (approved):
LGTM

Issue #524, opened 2023-09-06
todo: e2e testing


iuioiua:
Each e2e test is similar in format, and there are a lot of tests to cover. For these reasons, let's complete e2e in the following order, with each step meaning at least one new PR:
- [x] Achieve maximum coverage of e2e tests. Code duplication is acceptable at this stage.
- [x] Focus on code duplication and cleaning up the tests while maintaining coverage.

For any wanting to help, please give notice in this issue.

iuioiua:
For anyone wanting to work on this, the only work remaining for this is [/api/stripe-webhooks.ts](https://app.codecov.io/gh/denoland/saaskit/blob/main/routes%2Fapi%2Fstripe-webhooks.ts)

iuioiua:
We now have 100% coverage. The next step in `e2e_test.ts` will be deduplicating code and making tests consistent in structure, etc.

iuioiua:
@mbhrznr is taking care of the 2nd task.

PR #549, opened 2023-09-09
chore: add e2e tests for `GET /account/manage` and `GET /account/upgrade`

revgum:
First contribution towards improving e2e tests in #524

### DIFF
--- a/e2e_test.ts
+++ b/e2e_test.ts
@@ -67,6 +67,71 @@ Deno.test("[e2e] GET /account", async () => {
   assertEquals(resp.status, 303);
 });
 
+Deno.test("[e2e] GET /account/manage", async (test) => {
+  const url = "http://localhost/account/manage";
+
+  const resp1 = await handler(new Request(url));
+  assertFalse(resp1.ok);
+  assertFalse(resp1.body);
+  assertEquals(resp1.headers.get("location"), "/signin");
+  assertEquals(resp1.status, 303);
+
+  await test.step("returns HTTP 404 Not Found response if user does not have a stripeCustomerId", async () => {
+    const user = genNewUser();
+    await createUser({ ...user, stripeCustomerId: undefined });
+    const resp = await handler(
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+});
+
+Deno.test("[e2e] GET /account/upgrade", async (test) => {
+  const url = "http://localhost/account/upgrade";
+
+  const resp1 = await handler(new Request(url));
+  assertFalse(resp1.ok);
+  assertFalse(resp1.body);
+  assertEquals(resp1.headers.get("location"), "/signin");
+  assertEquals(resp1.status, 303);
+
+  const user = genNewUser();
+  await createUser(user);
+
+  await test.step("returns HTTP 500 Internal Server Error response if stripe premium plan price id is not set", async () => {
+    Deno.env.set("STRIPE_SECRET_KEY", crypto.randomUUID());
+    Deno.env.delete("STRIPE_PREMIUM_PLAN_PRICE_ID");
+    const resp = await handler(
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.InternalServerError);
+  });
+
+  await test.step("returns HTTP 404 Not Found response if stripe is not enabled", async () => {
+    Deno.env.set("STRIPE_PREMIUM_PLAN_PRICE_ID", crypto.randomUUID());
+    Deno.env.delete("STRIPE_SECRET_KEY");
+    const resp = await handler(
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+
+  Deno.env.delete("STRIPE_SECRET_KEY");
+  Deno.env.delete("STRIPE_PREMIUM_PLAN_PRICE_ID");
+});
+
 Deno.test("[e2e] GET /callback", async () => {
   const resp = await handler(
     new Request("http://localhost/callback"),
--- a/e2e_test.ts
+++ b/e2e_test.ts
@@ -82,7 +82,7 @@ Deno.test("[e2e] GET /account/manage", async (test) => {
     assertEquals(resp.status, 303);
   });
 
-  await test.step("returns HTTP 404 Not Found response if user does not have a stripeCustomerId", async () => {
+  await test.step("returns HTTP 404 Not Found response if the session user does not have a Stripe customer ID", async () => {
     const user = genNewUser();
     await createUser({ ...user, stripeCustomerId: undefined });
     const resp = await handler(
@@ -95,7 +95,7 @@ Deno.test("[e2e] GET /account/manage", async (test) => {
     assertEquals(resp.status, Status.NotFound);
   });
 
-  await test.step("returns redirect response to the url returned by stripe after creating a billing portal session", async () => {
+  await test.step("returns redirect response to the URL returned by Stripe after creating a billing portal session", async () => {
     const user = genNewUser();
     await createUser(user);
 
@@ -166,7 +166,7 @@ Deno.test("[e2e] GET /account/upgrade", async (test) => {
     assertEquals(resp.status, Status.NotFound);
   });
 
-  await test.step("returns HTTP 404 Not Found response if stripe returns a null url", async () => {
+  await test.step("returns HTTP 404 Not Found response if Stripe returns a `null` URL", async () => {
     Deno.env.set("STRIPE_PREMIUM_PLAN_PRICE_ID", crypto.randomUUID());
     Deno.env.set("STRIPE_SECRET_KEY", crypto.randomUUID());
 
@@ -190,7 +190,7 @@ Deno.test("[e2e] GET /account/upgrade", async (test) => {
     sessionsCreateStub.restore();
   });
 
-  await test.step("returns redirect response to the url returned by stripe after creating a checkout session", async () => {
+  await test.step("returns redirect response to the URL returned by Stripe after creating a checkout session", async () => {
     const priceId = crypto.randomUUID();
     Deno.env.set("STRIPE_PREMIUM_PLAN_PRICE_ID", priceId);
     Deno.env.set("STRIPE_SECRET_KEY", crypto.randomUUID());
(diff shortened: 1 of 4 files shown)


revgum:
@iuioiua I was just familiarizing myself with the saaskit and figured I could take a crack at helping out.

The refactoring here was intended to follow the intended behavior (HTTP 404). The original implementation throws, but the middleware upstream is just rethrowing which ends up being a HTTP 500. (https://github.com/denoland/saaskit/blob/main/middleware/errors.ts#L15)

revgum:
This failed test is also very confusing because it throws HTTP 500 locally .. 😕

routes/account/manage.ts
```diff
@@ -3,21 +3,14 @@ import type { RouteContext } from "$fresh/server.ts";
...
-  if (stripe === undefined) throw createHttpError(Status.NotFound);
-
   const { sessionUser } = ctx.state;
-  if (sessionUser.stripeCustomerId === undefined) {
-    throw createHttpError(
-      Status.NotFound,
-      "User does not have a Stripe customer ID",
-    );
+  if (stripe === undefined || sessionUser.stripeCustomerId === undefined) {
+    return ctx.renderNotFound();
```
revgum, line 13: This is a page, ultimately redirecting, so error handling by the middleware is different than the error handling for APIs. Without rethinking the middleware I figured I would propose an easy general 404.

e2e_test.ts
```diff
@@ -67,6 +67,71 @@ Deno.test("[e2e] GET /account", async () => {
...
+    const resp = await handler(
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+});
```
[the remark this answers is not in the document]
iuioiua: We could also add a test that tests whether the entire handler executes correctly. This would mean creating a Stripe customer. This test step could be ignored if `isStripeEnabled()` is false. This might be tedious, so I'm happy to tackle this in a proceeding PR.

e2e_test.ts
```diff
@@ -67,6 +67,71 @@ Deno.test("[e2e] GET /account", async () => {
...
+  const resp1 = await handler(new Request(url));
+  assertFalse(resp1.ok);
+  assertFalse(resp1.body);
+  assertEquals(resp1.headers.get("location"), "/signin");
+  assertEquals(resp1.status, 303);
+
+  const user = genNewUser();
+  await createUser(user);
+
+  await test.step("returns HTTP 500 Internal Server Error response if stripe premium plan price id is not set", async () => {
```
iuioiua, line 138:
```suggestion
  await test.step("returns HTTP 500 Internal Server Error response if the `STRIPE_PREMIUM_PLAN_PRICE_ID` environment variable is not set", async () => {
```

e2e_test.ts
```diff
@@ -67,6 +67,71 @@ Deno.test("[e2e] GET /account", async () => {
...
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.InternalServerError);
+  });
+
+  await test.step("returns HTTP 404 Not Found response if stripe is not enabled", async () => {
```
iuioiua, line 151:
```suggestion
  await test.step("returns HTTP 404 Not Found response if Stripe is disabled", async () => {
```

e2e_test.ts
```diff
@@ -67,6 +67,71 @@ Deno.test("[e2e] GET /account", async () => {
...
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+
+  Deno.env.delete("STRIPE_SECRET_KEY");
+  Deno.env.delete("STRIPE_PREMIUM_PLAN_PRICE_ID");
```
iuioiua, line 227:
```suggestion
```
These may not be needed. Is that correct?

e2e_test.ts
```diff
@@ -26,6 +11,24 @@ import {
...
+  assert,
+  assertArrayIncludes,
+  assertEquals,
+  assertFalse,
+  assertInstanceOf,
+  assertNotEquals,
+  assertStringIncludes,
+} from "std/assert/mod.ts";
+import { assertSpyCall, resolvesNext, stub } from "std/testing/mock.ts";
+import Stripe from "stripe";
```
revgum, line 31: This was just a combination of importing stripe and Stripe (for types) along with an import sort.

routes/account/upgrade.ts
```diff
@@ -26,13 +27,15 @@ export default async function AccountUpgradePage(
...
-        price: STRIPE_PREMIUM_PLAN_PRICE_ID,
+        price: stripePremiumPlanPriceId,
         quantity: 1,
       },
     ],
     mode: "subscription",
   });
-  if (url === null) throw createHttpError(Status.NotFound);
+  if (url === null) {
+    return ctx.renderNotFound();
```
revgum, line 37: Again with a basic 404 since this is a page route instead of an api route.

routes/account/manage.ts
iuioiua: Yeah, I was thinking about this too. I think there's no point in throwing `Status.NotFound` errors in the web page section of the site. I'll look into this separately, as I'll be creating an error-handling plugin.

routes/account/manage.ts
```diff
@@ -3,21 +3,14 @@ import type { RouteContext } from "$fresh/server.ts";
...
 ) {
-  if (stripe === undefined) throw createHttpError(Status.NotFound);
-
   const { sessionUser } = ctx.state;
-  if (sessionUser.stripeCustomerId === undefined) {
-    throw createHttpError(
-      Status.NotFound,
-      "User does not have a Stripe customer ID",
-    );
+  if (stripe === undefined || sessionUser.stripeCustomerId === undefined) {
```
iuioiua, line 12:
```suggestion
  if (!isStripeEnabled() || sessionUser.stripeCustomerId === undefined) {
```

iuioiua (changes requested):
Nicely done so far!

e2e_test.ts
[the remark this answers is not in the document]
revgum: > We should add a test step that tests for when `stripe.billingPortal.sessions.create()` is executed. It'll throw, but testing for that expected behaviour will be good.

Would you be concerned at all with having the tests hit Stripe and failing an authentication check.. just from a resource perspective on Stripes side. I doubt it would rise to the level of rate-limiting or whatever, but it made me wonder.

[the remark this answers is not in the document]
iuioiua: Actually, no longer needed, thanks to the mocking tests you added.

e2e_test.ts
```diff
@@ -67,6 +70,165 @@ Deno.test("[e2e] GET /account", async () => {
...
+
+  await test.step("returns redirect response if the session user is not signed in", async () => {
+    const resp = await handler(new Request(url));
+    assertFalse(resp.ok);
+    assertFalse(resp.body);
+    assertEquals(resp.headers.get("location"), "/signin");
+    assertEquals(resp.status, 303);
+  });
+
+  await test.step("returns HTTP 404 Not Found response if user does not have a stripeCustomerId", async () => {
```
iuioiua, line 85:
```suggestion
  await test.step("returns HTTP 404 Not Found response if the session user does not have a Stripe customer ID", async () => {
```

e2e_test.ts
```diff
@@ -67,6 +70,165 @@ Deno.test("[e2e] GET /account", async () => {
...
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+
+  await test.step("returns redirect response to the url returned by stripe after creating a billing portal session", async () => {
```
iuioiua, line 98:
```suggestion
  await test.step("returns redirect response to the URL returned by Stripe after creating a billing portal session", async () => {
```

routes/account/upgrade.ts
```diff
@@ -1,21 +1,22 @@
...
 
 export default async function AccountUpgradePage(
   _req: Request,
   ctx: RouteContext<undefined, SignedInState>,
 ) {
-  if (!isStripeEnabled()) throw createHttpError(Status.NotFound);
-  if (STRIPE_PREMIUM_PLAN_PRICE_ID === undefined) {
+  if (!isStripeEnabled()) {
+    return ctx.renderNotFound();
+  }
```
iuioiua, line 17:
```suggestion
  if (!isStripeEnabled()) return ctx.renderNotFound();
```

routes/account/upgrade.ts
```diff
@@ -26,13 +27,15 @@ export default async function AccountUpgradePage(
...
+        price: stripePremiumPlanPriceId,
         quantity: 1,
       },
     ],
     mode: "subscription",
   });
-  if (url === null) throw createHttpError(Status.NotFound);
+  if (url === null) {
+    return ctx.renderNotFound();
+  }
```
iuioiua, line 38:
```suggestion
  if (url === null) return ctx.renderNotFound();
```

e2e_test.ts
```diff
@@ -67,6 +70,165 @@ Deno.test("[e2e] GET /account", async () => {
...
+      new Request(url, {
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+  });
+
+  await test.step("returns HTTP 404 Not Found response if stripe returns a null url", async () => {
```
iuioiua, line 169:
```suggestion
  await test.step("returns HTTP 404 Not Found response if Stripe returns a `null` URL", async () => {
```

e2e_test.ts
```diff
@@ -67,6 +70,165 @@ Deno.test("[e2e] GET /account", async () => {
...
+        headers: { cookie: "site-session=" + user.sessionId },
+      }),
+    );
+
+    assertFalse(resp.ok);
+    assertEquals(resp.status, Status.NotFound);
+    sessionsCreateStub.restore();
+  });
+
+  await test.step("returns redirect response to the url returned by stripe after creating a checkout session", async () => {
```
iuioiua, line 193:
```suggestion
  await test.step("returns redirect response to the URL returned by Stripe after creating a checkout session", async () => {
```

iuioiua (changes requested):

iuioiua (approved):
Again, very nicely done. Thank you, @revgum! Let me know if there are any other issues that interest you.

revgum:
> Again, very nicely done. Thank you, @revgum! Let me know if there are any other issues that interest you.

Fun times, thanks for the quick turn on this! I just added you as a friend on Discord in case that's a better channel for async conversation. Saaskit is a great project! 

I've got a little free time, I'll probably just spin up another PR with some more test coverage!

PR #555, opened 2023-09-10
chore: adds e2e tests for `POST /api/comments`

revgum:
Adding more e2e coverage for #524

### DIFF
--- a/e2e_test.ts
+++ b/e2e_test.ts
@@ -28,7 +28,14 @@ import {
   assertNotEquals,
   assertStringIncludes,
 } from "std/assert/mod.ts";
-import { assertSpyCall, resolvesNext, stub } from "std/testing/mock.ts";
+import {
+  assertSpyCall,
+  assertSpyCallAsync,
+  assertSpyCalls,
+  resolvesNext,
+  spy,
+  stub,
+} from "std/testing/mock.ts";
 import Stripe from "stripe";
 import options from "./fresh.config.ts";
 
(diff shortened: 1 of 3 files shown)


e2e_test.ts
```diff
@@ -624,3 +631,101 @@ Deno.test("[e2e] GET /notifications/[id]", async (test) => {
...
+      new Request(url, {
+        method: "POST",
+        headers: { cookie: "site-session=" + user.sessionId },
+        body: formData,
+      }),
+    );
+
+    assertEquals(resp.status, Status.SeeOther);
+    assertSpyCallAsync(createCommentSpy, 0);
+    assertSpyCalls(createNotificationSpy, 0);
```
revgum, line 710: Calling this out, this is a way to detect that a spy is not called by testing the number of calls that were made to it.. whereas `assertSpyCall` and `assertSpyCallAsync` is actually testing that an `index` call is being made. 

Since a new `ulid()` is created when a comment is built, I wasn't able to find a way around checking the `args` provided to the spy, just that it was called.

iuioiua: I used `assertSpyCalls()` in all instances for `createComment()` and `createNotification()`. Frankly, the code looks correct to me. However, tests are failing, saying that both test steps never call `createComment()` and `createNotification()`, which seems incorrect. Am I missing something here? @kt3k, does this look right to you?

revgum: Odd.. yea, I tested a bit more to validate my misunderstanding here.. for instance;

`createCommentSpy.calls` has nothing..even though `assertSpyCallAsync` seems like it is "working", and now I couldn't be sure... Dang, I'll dig in a little bit to see what's wrong here.

revgum: I think part of the problem here is that `createComment` (for instance) is also imported into `e2e_test.ts` to facilitate test setup. I tried to spy on `kv.atomic` and that spy wasn't working in the context of this test either. Hrm!

revgum: Rethinking the test and using a spy here.. I added a commit that replaces the spy on these two tests with queries and existence checks. Let me know what you think?

iuioiua (approved):
I made a few tweaks, and now, LGTM! Thanks again, Josh!

iuioiua:
Also, yes, it's a good idea to avoid mocking as much as possible, only using it as a last resort, IMO.

PR #559, opened 2023-09-10
fix: avoid showing "No items found" stats in `<ItemsList />`

### LINKED ISSUE #472: todo: avoid showing "No items found" state on home and user page
When the page is loading, this "No items found" state is shown, which is misleading. We can instead just show nothing while loading.

https://github.com/denoland/saaskit/assets/29347852/a5a010a2-671c-43d4-bea5-bc96cf0661b0

roberto-morado:
fixes #472 

@iuioiua this is as simple as I got, without being too intrusive; I personally like it; the problem was a bug regarding the state change of the variable `isLoadingSig`, I was able to avoid it by initializing the variable as `undefined` and bypassing it with the conditional rendering not sure if is worth taking another look at it some other time.

Note: The simplicity of this PR does not reflect at all the time that took me to get there — just for the record.

### DIFF
--- a/islands/ItemsList.tsx
+++ b/islands/ItemsList.tsx
@@ -75,22 +75,24 @@ export default function ItemsList(
       .finally(() => loadMoreItems());
   }, []);
 
+  if (isLoadingSig.value === undefined || isLoadingSig.value === true) {
+    return <>█</>;
+  }
+
   return (
     <div>
-      {isLoadingSig.value !== undefined && !isLoadingSig.value
-        ? itemsSig.value.length > 0
-          ? itemsSig.value.map((item, id) => {
-            return (
-              <ItemSummary
-                key={item.id}
-                item={item}
-                isVoted={itemsAreVotedSig.value[id]}
-                isSignedIn={props.isSignedIn}
-              />
-            );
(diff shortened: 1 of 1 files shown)


islands/ItemsList.tsx
```diff
@@ -77,18 +77,20 @@ export default function ItemsList(
...
+              <ItemSummary
+                key={item.id}
+                item={item}
+                isVoted={itemsAreVotedSig.value[id]}
+                isSignedIn={props.isSignedIn}
+              />
+            );
+          })
+          : <EmptyItemsList />
+        : "█"}
```
iuioiua, line 93: Can you make this a little more readable, please? I think we should avoid nesting ternary operators.

roberto-morado: Sure thing

roberto-morado:
@iuioiua man! how right you were. This reads way better.

roberto-morado:
After seeing this PR I cannot help but wonder what was I doing this whole time? 🤔

iuioiua (approved):
Made a couple of tweaks, and now LGTM. Thank you.

PR #583, opened 2023-09-12
fix: prevent homepage from crashing

mbhrznr:
currently https://hunt.deno.land/ crashes with the following error in the console:

Bildschirm­foto 2023-09-13 um 00 17 44

this seems to happen whenever any of the items still uses the old data model w/ `uuid` and `createdAt`, rather than `ulid`.
this pr adds a quick and naive `try/catch` block to prevent the page items list from crashing.

however the underlying issue should be handled via a db migration on prod's database.


iuioiua (approved):
Let's try this out! Thank you

PR #585, opened 2023-09-13
fix:redirect to home page after project created.

harshmangalam:


iuioiua:
@harshmangalam, this seemed like a valid fix.

harshmangalam:
Thanks @iuioiua I have make PR here https://github.com/denoland/saaskit/pull/587

Issue #588, opened 2023-09-13
Bug report


digioak:
**Describe the bug** A clear and concise description of what the bug is.

In `db_seed.ts` task, increasing `TOP_STORIES_COUNT` to larger value (for example, 100), may cause `createUser()` to fail due to duplicate user logins.

**To Reproduce** Steps to reproduce the behavior:

1. Increate `TOP_STORIES_COUNT` to `100`
2. `deno task db:seed`

**Expected behavior** A clear and concise description of what you expected to
happen.

New projects and associated users seed without errors.

**Screenshots** If applicable, add screenshots to help explain your problem.

**Desktop (please complete the following information):**

- OS: macOS
- Browser: n/a
- Version n/a

**Smartphone (please complete the following information):**

- Device: n/a
- OS: macOS
- Browser: n/a
- Version n/a

**Additional context** Add any other context about the problem here.

iuioiua:
Completed in #589

PR #589, opened 2023-09-13
fix: prevent duplicate users in seed task

digioak:
See https://github.com/denoland/saaskit/issues/588

### DIFF
--- a/tasks/db_seed.ts
+++ b/tasks/db_seed.ts
@@ -19,7 +19,7 @@ interface Story {
 }
 
 const resp = await fetch(API_TOP_STORIES_URL);
-const allTopStories = await resp.json() as number[];
+const allTopStories = (await resp.json()) as number[];
 const topStories = allTopStories.slice(0, TOP_STORIES_COUNT);
 const storiesPromises = [];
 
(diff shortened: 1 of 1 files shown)


tasks/db_seed.ts
```diff
@@ -40,17 +40,19 @@ const items = stories.map(({ by: userLogin, title, url, score, time }) => ({
   createdAt: new Date(time * 1000),
 })).filter(({ url }) => url);
 
+const users = Array.from(new Set(items.map((user) => user.userLogin)));
```
iuioiua, line 43:
```suggestion
const users = new Set(items.map((user) => user.userLogin));
```
`Set` is iterable. `Array.from()` is not needed 🙂

iuioiua (changes requested):

tasks/db_seed.ts
digioak: > `Set` is iterable. `Array.from()` is not needed 🙂

Very true! Got confused because of mapping and filtering earlier on. Fixed.

iuioiua:
Please run `deno fmt`

digioak:
> Please run `deno fmt`

Had to disable my local prettier and eslint.

iuioiua (approved):
LGTM. Thank you

Issue #591, opened 2023-09-14
todo: improve security of HTTP headers


iuioiua:
HTTP headers have much room for improvement, in terms of security. This should be fixed by a plugin with reasonable settings. Report: https://observatory.mozilla.org/analyze/hunt.deno.land

Jabolol:
I'd like to tackle this issue, can I get it assigned? Thanks!

Jabolol:
The report was at `20/100`, and `F` grade. I've managed to get a `80/100`, that is a `B+`. I've encountered mayor obstacles while trying to implement the last suggestion, the [`Content-Security-Policy`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) header. Implementing this would achieve a perfect score.

## The problem

Content Security Policy (CSP from now on) is a security feature used in web applications to mitigate certain types of attacks, such as cross-site scripting (XSS) and data injection. Mozilla's Observatory requires the values `script-src` and `style-src` of the CSP header to not be set as `unsafe-inline`. Due to the nature of fresh (scripts and styles being injected and inlined at build time), there is no practical way of overcoming this limitation without modifying fresh itself.

## Possible solutions

1. Compute the `sha256`/`sha384` hash at build step for every `script` tag at the `fresh` level itself and set it in the [`integrity`](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity), then set the `script-src` CSP value to [`strict-dynamic`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src#strict-dynamic).
2. Create a middleware handler that checks if the file is `internal` (through `ctx.destination`) and a JavaScript file, compute the `sha256`/`sha384` hash on the fly and set it in the CSP header.

## Thoughts

For the moment, an `80/100` score is good enough, I suggest to implement one of the aforementioned after the `v1` release, unless someone comes up with a better solution. The second suggestion is slow for big files, and that would take a considerable toll on lighthouse score, making it inviable.

Here's the Observatory report for my PR (#592): https://observatory.mozilla.org/analyze/saaskit-staging.deno.dev

marvinhagemeister:
@Jabolol The CSP issue will be fixed in Fresh itself. Already did so for inline scripts (see https://github.com/denoland/fresh/pull/1772), and inline styles are next on my agenda.

iuioiua:
[...]

PR #592, opened 2023-09-16
feat: harden headers security

Jabolol:
Work in progress. See #591 for the relevant discussion.


middleware/headers.ts
```diff
@@ -0,0 +1,36 @@
...
+  req: Request,
+  ctx: MiddlewareHandlerContext,
+) {
+  const response = await ctx.next();
+  const path = new URL(req.url).pathname.split("/")[1] || "/";
+  const blacklist = [/api/];
+
+  if (blacklist.some((regex) => regex.test(path))) {
+    return response;
+  }
```
iuioiua, line 14:
```suggestion
  if (
    ctx.destination !== "route" || new URL(req.url).pathname.startsWith("/api")
  ) return await ctx.next();
  
  const response = await ctx.next();
```

middleware/headers.ts
```diff
@@ -0,0 +1,36 @@
+// Copyright 2023 the Deno authors. All rights reserved. MIT license.
+import type { MiddlewareHandlerContext } from "$fresh/server.ts";
+
+export async function hardenHeaders(
```
iuioiua, line 4: We can move this to a plugin. I want to remove the `/middleware` folder in favour of the `/plugins` folder. The plugin can have the name `security-headers`.

iuioiua (changes requested):
Good start! See if [this](https://fresh.deno.dev/docs/examples/using-csp) is valuable to you in this PR.

Jabolol:
> Good start! See if [this](https://fresh.deno.dev/docs/examples/using-csp) is valuable to you in this PR.

Oh I completely missed that section of the docs! It’s most useful, thanks for the heads up. I’ll implement the changes requested ASAP.

middleware/headers.ts
Jabolol: I was using that approach because I wanted to apply that middleware to all routes. I want to avoid adding all the routes manually as in this snippet:

```ts
// ... imports
const middleware = { handler: hardenHeaders };
export default {
  name: "security-headers",
  middlewares: [
    {
      path: "/",
      middleware,
    },
    // ... add routes manually here
  ],
} as Plugin<State>;
```

Is there an alternative to this?

iuioiua: Yep, that's the right way to do it 👍🏾

Jabolol:
Once https://github.com/denoland/fresh/pull/1787 lands we will be able to achieve an `A+` score

iuioiua:
> Once [denoland/fresh#1787](https://github.com/denoland/fresh/pull/1787) lands we will be able to achieve an `A+` score

Let's add a `@todo` in this PR. I'd like to merge this PR sooner rather than depend on that PR.

plugins/security_headers.ts
```diff
@@ -0,0 +1,36 @@
...
+      middleware: {
+        handler: async (req, ctx) => {
+          if (
+            ctx.destination !== "route" ||
+            new URL(req.url).pathname.startsWith("/api")
+          ) return await ctx.next();
+
+          const response = await ctx.next();
+
+          // TODO(Jabolol): Implement Content-Security-Policy
```
iuioiua, line 18: Well, we can reference that PR here.

iuioiua (changes requested):
Looking better! Can you please add a test?

Jabolol:
> Looking better! Can you please add a test?

Done! Comment improved and added a test case. Let me know if I should add something else

iuioiua:
Test logic looks good! However, are we able to do this in a `/plugins/security_headers_test.ts` file?

Jabolol:
> Test logic looks good! However, are we able to do this in a `/plugins/security_headers_test.ts` file?

Yeah, that's better indeed. Moved it.

iuioiua:
Is this ready for review?

iuioiua:
Closes #591.

iuioiua (approved):
I'm going to assume this PR is ready to go. I tested it on my local machine, and it works without hiccups! LGTM! Thank you, @Jabolol.

PR #596, opened 2023-09-21
tests(e2e): harmonize e2e to make use of custom assertions

mbhrznr:
working towards #524.

adds a couple custom assertion fns to have a predictable set to work with.


e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
  */
 const handler = await createHandler(manifest, options);
 
-function assertResponseNotFound(resp: Response) {
+function assertResponseAccepted(resp: Response) {
+  assert(resp.ok);
+  assertEquals(resp.body, null);
+  assertEquals(resp.status, Status.Accepted);
+}
```
iuioiua, line 41: I don't think this, and other assertions that solely check response status, are needed. Instead, one can just do `assertEquals(resp.status, status)`. Custom assertions are not needed.

Also, checking the body and `ok` are not needed. I'd go as far as to say that if `resp.status` is checked, `resp.ok` never needs to be checked.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+
+function assertResponseBadRequest(resp: Response) {
```
iuioiua, line 43: Ditto

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
   assertFalse(resp.ok);
-  assertEquals(resp.status, Status.NotFound);
+  assertEquals(
+    resp.headers.get("content-type"),
+    "text/plain;charset=UTF-8",
+  );
+  assertEquals(resp.status, Status.BadRequest);
+}
+
+function assertResponseCreated(resp: Response) {
```
iuioiua, line 52: Ditto

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+}
+
+function assertResponseHtml(resp: Response) {
+  assert(resp.ok);
+  assertInstanceOf(resp.body, ReadableStream);
+  assertEquals(
+    resp.headers.get("content-type"),
+    "text/html; charset=utf-8",
+  );
+  assertEquals(resp.status, Status.OK);
```
iuioiua, line 65:
```suggestion
function assertHtml(resp: Response) {
  assertInstanceOf(resp.body, ReadableStream);
  assertEquals(
    resp.headers.get("content-type"),
    "text/html; charset=utf-8",
  );
```
This does a little more than it needs to.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
 }
 
 function assertResponseJson(resp: Response) {
   assert(resp.ok);
   assertNotEquals(resp.body, null);
   assertEquals(resp.headers.get("content-type"), "application/json");
+  assertEquals(resp.status, Status.OK);
 }
```
iuioiua, line 73:
```suggestion
function assertJson(resp: Response) {
  assertInstanceOf(resp.body, ReadableStream);
  assertEquals(resp.headers.get("content-type"), "application/json");
}
```
Ditto. This does a little more than needed.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
 
-Deno.test("[e2e] GET /", async () => {
-  const resp = await handler(new Request("http://localhost"));
+function assertResponseNoContent(resp: Response) {
+  assert(resp.ok);
+  assertEquals(resp.body, null);
+  assertEquals(resp.status, Status.NoContent);
+}
```
iuioiua, line 79: Ditto. Not needed.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
 
+function assertResponseNotFound(resp: Response, header: string) {
```
iuioiua, line 81: Ditto. Not needed.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+function assertResponseRedirect(
+  resp: Response,
+  location: string,
+  status: Status,
+) {
+  assertFalse(resp.ok);
+  assertFalse(resp.body);
+  assertStringIncludes(resp.headers.get("location")!, location);
+  assertEquals(resp.status, status);
+}
```
iuioiua, line 99:
```suggestion
function assertRedirect(response: Response, location?: string) {
  assert(isRedirectStatus(response.status));
  if (location !== undefined) {
    assertEquals(response.headers.get("location"), location);
  } else {
    assert(response.headers.has("location"));
  }
}
```
We can use [`isRedirectStatus()`](https://deno.land/std@0.202.0/http/mod.ts?s=isRedirectStatus).

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+
+function assertResponseServerError(resp: Response, header: string) {
```
iuioiua, line 101: Ditto. Not needed.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+}
+
+function assertResponseUnauthorized(resp: Response) {
+  assertFalse(resp.ok);
+  assertEquals(
+    resp.headers.get("content-type"),
+    "text/plain;charset=UTF-8",
+  );
+  assertEquals(resp.status, Status.Unauthorized);
+}
```
iuioiua, line 118: Ditto. Not needed.

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
+
+function assertResponseXml(resp: Response) {
   assert(resp.ok);
```
iuioiua, line 121:
```suggestion
```

e2e_test.ts
```diff
@@ -34,55 +34,119 @@ import options from "./fresh.config.ts";
...
   assertInstanceOf(resp.body, ReadableStream);
   assertEquals(
     resp.headers.get("content-type"),
-    "text/html; charset=utf-8",
+    "application/atom+xml; charset=utf-8",
   );
-  assertEquals(resp.status, 200);
+  assertEquals(resp.status, Status.OK);
```
iuioiua, line 127:
```suggestion
```

iuioiua (changes requested):
Ok, this might take a bit of back-and-forth. Let's make the aim of this PR to establish a simple mental framework for writing these e2e tests.

One thing for sure is that each test should assert its `response.status`. Let's call that rule one. Related to that is if `response.status` is asserted, we never need to check `response.ok`, because the value of `response.ok` is calculated based on `response.status`.

Secondly, I think custom assertions are only needed if the assertion checks more than 1 property of the response. E.g. `responseJson()` checks that the body is an instance of `ReadableStream` and the `content-type` header is for JSON.

e2e_test.ts
mbhrznr: midway through the first assertions i got carried away.
having the defined ruleset helped a lot!

e2e_test.ts
mbhrznr: applied as suggested!

e2e_test.ts
mbhrznr: applied as suggested!

e2e_test.ts
mbhrznr: `std` is such a beast.

e2e_test.ts
```diff
@@ -34,40 +33,47 @@ import options from "./fresh.config.ts";
...
     new Request("http://localhost/callback"),
   );
 
-  assertFalse(resp.ok);
-  assertInstanceOf(resp.body, ReadableStream);
-  assertEquals(
-    resp.headers.get("content-type"),
-    "text/html; charset=utf-8",
-  );
   assertEquals(resp.status, 500);
```
iuioiua, line 77: Let's make use of the `Status` enum where we can. It let's the reader know what the status code means.

e2e_test.ts
```diff
@@ -34,40 +33,47 @@ import options from "./fresh.config.ts";
...
-    "text/html; charset=utf-8",
+    "application/atom+xml; charset=utf-8",
   );
-  assertEquals(resp.status, 200);
+}
+
+Deno.test("[e2e] GET /", async () => {
+  const resp = await handler(new Request("http://localhost"));
+
+  assertHtml(resp);
```
iuioiua, line 69: Let's not forget to always check the response status first.

iuioiua (changes requested):
I left some comments that should be checked and have been put in place elsewhere in the e2e tests. We should also have a comment block at the top of the file stating the ruleset.

e2e_test.ts
iuioiua: Sorry, to be clear, the status assertion should appear before any custom assertion. The custom assertion should not include the status check unless required. E.g. A status check in `assertRedirect()` makes sense, but a status check in `assertHtml()` is beyond the scope of the assertion. I hope this makes sense.

iuioiua (changes requested):
Looking better! However, we should be careful to only include status checks within custom assertions only when necessary.

Also, looking at things now, I think that we'll be fine without a single, multipurpose custom assertion. It's looking pretty good now. WDYT?

e2e_test.ts
mbhrznr: sorry about the misunderstanding!
extracted the status checks on all custom assertions but `assertRedirect`.

mbhrznr:
> Also, looking at things now, I think that we'll be fine without a single, multipurpose custom assertion. It's looking pretty good now. WDYT?

fully agreeing. really like how it turned out w/ our defined rules!

iuioiua (approved):
I've made the tweaks by implementing my suggestion of moving status checks outside of custom assertions. With a few other tweaks, LGTM! Thank you for this.

PR #638, opened 2023-10-30
refactor: remove use of `std/http/http_errors`

iuioiua:
It appears `std/http/http_errors` will be getting removed (see [here](https://github.com/denoland/deno_std/pull/3736)). Removing this dependency actually makes things cleaner.


kt3k (approved):
LGTM 👍

PR #639, opened 2023-11-02
refactor: use `Array.fromAsync()`

iuioiua:
[Deno 1.38](https://deno.com/blog/v1.38#v8-120) now supports [`Array.fromAsync()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fromAsync).


marvinhagemeister:
This is currently causing 500 errors in prod. Whilst `Array.fromAsync` is supported in Deno CLI, it's not yet supported in Deploy.

Issue #644, opened 2023-12-03
todo: upgrade to Fresh 1.6


iuioiua:
SaaSKit needs to be upgraded to [Fresh 1.6, which was just released](https://deno.com/blog/fresh-1.6). We should also switch from using Twind to Tailwind CSS (guide [here](https://fresh.deno.dev/docs/examples/migrating-to-tailwind)). Once this is done, I'll publish a new minor version of SaaSKit.

gabobaxx:
Hi. Why will you change from twind to tailwind? 
I lost track of SassKit a little 🥺

iuioiua:
Hi @gabobaxx, give [this](https://deno.com/blog/fresh-1.6#first-class-tailwind-css-plugin) and [this](https://fresh.deno.dev/docs/examples/migrating-to-tailwind#frequently-asked-questions-faq) a read. I think they should answer your questions 🙂

PR #646, opened 2023-12-06
chore: upgrade to Fresh 1.6.1 and std 0.208.0

deer:
closes https://github.com/denoland/saaskit/issues/644

I see the guidelines have changed since I last contributed (thanks github!) and that caused me to look for an issue. Fortunately there was the above, but I see you want tailwind as well. I'll open this as a draft for now.


deer:
@iuioiua, this is ready to go now. Although you might consider some style tests somewhere. I almost botched this by forgetting to add the `styles.css` file. Of course the site was unstyled, but all the tests were passing.

iuioiua:
Pretty much LGTM! I'll give this a spin tomorrow before approving.

As for the styles test, let's create an issue to fix after we land this PR.

marvinhagemeister (dismissed):

deer:
Thanks guys. I'll try again on Friday, first ensuring a near pixel perfect match. I guess I need to familiarize myself with the app's functionality 😅

iuioiua:
Hm... I think Tailwind can't recognise classes if stored in variables. The fix may be as simple as removing "class variables", like `LINK_STYLES`, and just using the classes directly. We can also create custom classes and reuse those.

deer:
Still fighting with active links, but it's overall much better now.

routes/welcome.tsx
```diff
@@ -5,7 +5,7 @@ import { isGitHubSetup } from "@/utils/github.ts";
 
 function SetupInstruction() {
   return (
-    <div class="bg-green-50 dark:(bg-gray-900 border border-green-800) rounded-xl max-w-screen-sm mx-auto p-8 space-y-2">
+    <div class="bg-green-50 dark:bg-gray-900 dark:border dark:border-green-800 rounded-xl max-w-screen-sm mx-auto p-8 space-y-2">
```
iuioiua, line 8: Are grouped classes no longer working?

deer: https://fresh.deno.dev/docs/examples/migrating-to-tailwind

> Tailwind CSS doesn’t support the grouping syntax from twind: text(lg uppercase gray-100). These need to be rewritten to their expanded values like text-lg uppercase text-gray-100.

iuioiua: Oh, I wasn't aware that was a Twind thing. Cool.

deer:
* As per https://tailwindcss.com/docs/border-width, tailwind does not do `border-1`, so I had to define this in `tailwind.config.ts`.
* The solution to the link issue was to change the position of the exclamation mark.
* Somehow `tw-ring-opacity: 0.5;` was getting inserted into the old version automatically; I had to manually apply it here with `ring-opacity-50`.

iuioiua (approved):
LGTM! Awesome work, @deer! Thanks for this.

PR #659, opened 2024-01-21
chore: allow license check to be multiline

matthewford:
Currently, when you add your own license to modified files, it will complain.

This changes the regex so that it can be multiline.


iuioiua (approved):
This makes sense to have! LGTM. Thank you.

PR #660, opened 2024-01-24
feat: pluginify blog

deer:
@iuioiua, check it out. Converting it to a plugin is pretty straightforward. There's just this module resolution issue that I need to resolve. You can see the damage here, where I had to do this:
```diff
- import Head from "@/components/Head.tsx";
- import Share from "@/components/Share.tsx";
+ import Head from "../../../components/Head.tsx";
+ import Share from "../../../components/Share.tsx";
```


deer:
This is much more pleasing now that I got root imports working.

iuioiua:
Looking good! I made a couple of simplifications. What if we had a `plugins/blog` folder containing as much blog code as possible? We could measure it as a success if one simply deleted that folder and made minimal code changes to remove blog functionality altogether. Can you give that a try?

iuioiua:
1. If files are reasonably close to each other and are likely to stay in the same relative locations to each other, I think it's better to use relative paths. Just a bit easier to picture where things are relative to each other. Root imports work well when files aren't tightly related.
2. Yeah, I think `posts.ts` might be the only file needed. I'll have a second look when I can.

fresh.config.ts
```diff
@@ -7,6 +7,7 @@ import securityHeaders from "./plugins/security_headers.ts";
 import welcomePlugin from "./plugins/welcome.ts";
 import type { FreshConfig } from "$fresh/server.ts";
 import { ga4Plugin } from "https://deno.land/x/fresh_ga4@0.0.4/mod.ts";
+import { blog } from "./plugins/blog/blog.ts";
```
iuioiua, line 10:
```suggestion
import { blog } from "./plugins/blog/mod.ts";
```
Nit

routes/feed.ts
iuioiua, line 1: We should include RSS feed functionality into the blog plugin, as the blog is all that the RSS feed covers. That should simplify things a little.

iuioiua (changes requested):

plugins/blog/mod.ts
```diff
@@ -0,0 +1,24 @@
...
+      component: BlogIndex,
+    }, {
+      path: "/blog/[slug]",
+      component: BlogSlug,
+    }, {
+      path: "/feed",
+      component: Feed,
+    }],
+    location: import.meta.url,
+    projectLocation: toFileUrl(Deno.cwd()).href,
```
iuioiua, line 22: Not a blocker, but something interesting I observed is that Tailwind styling appeared to break when I used `projectLocation: Deno.cwd()`. Not sure why.

iuioiua (approved):
LGTM! I like this. Now let's see about that upstream PR.

iuioiua:
I might merge this PR once the upstream Fresh PR LGTM is so as not to be blocked by further reviews.

deer:
This doesn't actually work if someone else tries to use it outside of saaskit, which is the point of doing this 😅 

I'm trying to import this `mod.ts` and I'm getting errors like:
```
Watcher File change detected! Restarting!
error: Module not found "file:///Users/reed/code/deer/testproject/components/Head.tsx".
    at file:///Users/reed/code/denoland/saaskit/plugins/blog/routes/blog/[slug].tsx:5:18
Watcher Process failed. Restarting on file change...
```

The specifier `import Head from "@/components/Head.tsx";` from `[slug].tsx` is getting interpreted with my import map in my `testproject`.

Is there a way to stop this? Otherwise this seems to resolve it:
```diff
git diff
diff --git a/components/Head.tsx b/components/Head.tsx
index b624b09..5bb89cb 100644
--- a/components/Head.tsx
+++ b/components/Head.tsx
@@ -1,7 +1,7 @@
 // Copyright 2023-2024 the Deno authors. All rights reserved. MIT license.
 import { Head as _Head } from "$fresh/runtime.ts";
 import Meta, { type MetaProps } from "./Meta.tsx";
-import { SITE_DESCRIPTION, SITE_NAME } from "@/utils/constants.ts";
+import { SITE_DESCRIPTION, SITE_NAME } from "../utils/constants.ts";
 import { ComponentChildren } from "preact";

 export type HeadProps =
diff --git a/plugins/blog/mod.ts b/plugins/blog/mod.ts
index 18fdcad..17b9ff6 100644
--- a/plugins/blog/mod.ts
+++ b/plugins/blog/mod.ts
@@ -6,6 +6,8 @@ import Feed from "./routes/feed.ts";
 import { toFileUrl } from "std/path/to_file_url.ts";

 export function blog() {
   return {
     name: "blog",
     routes: [{
diff --git a/plugins/blog/routes/blog/[slug].tsx b/plugins/blog/routes/blog/[slug].tsx
index a910076..75ccf7a 100644
--- a/plugins/blog/routes/blog/[slug].tsx
+++ b/plugins/blog/routes/blog/[slug].tsx
@@ -2,8 +2,8 @@
 import { defineRoute } from "$fresh/server.ts";
 import { CSS, render } from "$gfm";
 import { getPost } from "../../utils/posts.ts";
-import Head from "@/components/Head.tsx";
-import Share from "@/components/Share.tsx";
+import Head from "../../../../components/Head.tsx";
+import Share from "../../../../components/Share.tsx";

 export default defineRoute(async (_req, ctx) => {
   const post = await getPost(ctx.params.slug);
diff --git a/plugins/blog/routes/blog/index.tsx b/plugins/blog/routes/blog/index.tsx
index 9bb1f3d..7dc7b18 100644
--- a/plugins/blog/routes/blog/index.tsx
+++ b/plugins/blog/routes/blog/index.tsx
@@ -1,7 +1,7 @@
 // Copyright 2023-2024 the Deno authors. All rights reserved. MIT license.
 import { defineRoute } from "$fresh/server.ts";
 import { getPosts, type Post } from "../../utils/posts.ts";
-import Head from "@/components/Head.tsx";
+import Head from "../../../../components/Head.tsx";

 function PostCard(props: Post) {
   return (
diff --git a/plugins/blog/routes/feed.ts b/plugins/blog/routes/feed.ts
index 7d19800..d7776a3 100644
--- a/plugins/blog/routes/feed.ts
+++ b/plugins/blog/routes/feed.ts
@@ -1,7 +1,7 @@
 // Copyright 2023-2024 the Deno authors. All rights reserved. MIT license.
 import { Feed } from "feed";
 import { getPosts } from "../utils/posts.ts";
-import { SITE_NAME } from "@/utils/constants.ts";
+import { SITE_NAME } from "../../../utils/constants.ts";
 import { defineRoute } from "$fresh/server.ts";

 const copyright = `Copyright ${new Date().getFullYear()} ${SITE_NAME}`;
```

Additionally I needed to add the following to my import map:
```json
    "std/": "https://deno.land/std@0.207.0/",
    "tabler_icons_tsx/": "https://deno.land/x/tabler_icons_tsx@0.0.4/tsx/",
    "$gfm": "https://deno.land/x/gfm@0.5.0/mod.ts",
    "feed": "npm:/feed@4.2.2",
```

If we're serious about this, then we need to take major steps to convert saaskit from an app to a framework. No more using import maps, but rather a `deps.ts` file. I think this also means no more `@/` imports. Otherwise we'll push the work to the users.

deer:
Additionally the styling of the blog is horrible, because there's no _app (at least in my project). So perhaps that should be the next plugin.

Or maybe we need a `SaaSKitPlugin` with a `SaaSKitOptions` bag. And then I could use it like this:
```ts
export default defineConfig({
  plugins: [
    tailwind(),
    ga4Plugin() as Plugin,
    saaskit({
      routes: {
        blog: true,
        feed: true,
        app: true,
        400: true,
        500: true,
      },
    }),
  ],
});
```

I would hesitate about merging this until we have another think about how this will work going forward.

iuioiua:
@deer, I've made a few tweaks to address your previous concerns. A couple of flaws should still be addressed, like using the `<Head />` component, which exists outside of the plugin. Nevertheless, let's merge this PR. Doing so allows us to test the upstream Fresh feature and provide better backing for adding it to Fresh itself. WDYT?

deno.json
```diff
@@ -20,8 +20,7 @@
   "compilerOptions": { "jsx": "react-jsx", "jsxImportSource": "preact" },
   "imports": {
     "@/": "./",
-    "$fresh/": "https://deno.land/x/fresh@1.6.3/",
-    "$gfm": "https://deno.land/x/gfm@0.2.5/mod.ts",
+    "$fresh/": "https://raw.githubusercontent.com/deer/fresh/2159_plugin_routes_tailwind/",
```
iuioiua, line 23: Are you able to commit-lock this import?

deno.json
```diff
@@ -30,9 +29,8 @@
     "tailwindcss": "npm:tailwindcss@3.4.1",
     "tailwindcss/": "npm:/tailwindcss@3.4.1/",
     "tailwindcss/plugin": "npm:/tailwindcss@3.4.1/plugin.js",
-    "std/": "https://deno.land/std@0.208.0/",
+    "$std/": "https://deno.land/std@0.208.0/",
```
iuioiua, line 32: This is the std import that comes with a Fresh install anyway.

deer: Very true.

deer:
@iuioiua, I've updated the pinned version to reference the most recent commit in the PR. This includes preact 10.19.6 and some other fixes from main.

More importantly it removes the offending `upNLevels` function. I've additionally updated the plugin here to match the pattern you specified in the PR. This has the benefit that the plugin is no longer dependent on `cwd`.

Please let me know if there's anything else to do here.

iuioiua:
Merging now as I think it's ready for the first pass. Let's iterate from here. Thank you very much, @deer and apologies for this taking so long.

PR #671, opened 2024-02-21
fix: hamburger menu styling on smaller screens

deer:
"Someone" broke the hamburger menu in #646.

This fixes the issue and cleans things up by making the CSS a bit more direct:
* I use https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-sibling-state to modify the `nav` directly.
* I add IDs to the hamburger and X components, and then use https://tailwindcss.com/docs/hover-focus-and-other-states#using-arbitrary-variants to directly target them, in combination with peer.

FYI @hashrock since this restores the great work you did in #374.

### DIFF
--- a/components/Header.tsx
+++ b/components/Header.tsx
@@ -21,10 +21,10 @@ export default function Header(props: HeaderProps) {
       <input
         type="checkbox"
         id="nav-toggle"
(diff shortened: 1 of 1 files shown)


iuioiua (approved):
LGTM! Thanks for fixing this 👍🏾

iuioiua:
I'll merge this as I'm sure it's correct. @hashrock, please feel free to review after merge.

PR #672, opened 2024-02-22
excluded dirs were included

inverted-capital:
the walk function passes the absolute path, not the relative one


iuioiua (approved):
LGTM. Thank you!

PR #674, opened 2024-04-13
refactor: kv_oauth

fdemir:
I was trying to use the google provider and noticed that kv_oauth was old. I've upgraded it and make it easy to add new providers etc. Feel free to give feedback.

### DIFF
--- a/deno.json
+++ b/deno.json
@@ -31,7 +31,6 @@
     "tailwindcss/plugin": "npm:/tailwindcss@3.4.1/plugin.js",
     "$std/": "https://deno.land/std@0.208.0/",
     "stripe": "npm:/stripe@13.5.0",
-    "kv_oauth/": "https://deno.land/x/deno_kv_oauth@v0.9.1/",
     "tabler_icons_tsx/": "https://deno.land/x/tabler_icons_tsx@0.0.4/tsx/",
     "fresh_charts/": "https://deno.land/x/fresh_charts@0.3.1/"
   },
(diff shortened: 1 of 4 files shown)


iuioiua:
Thank you for the PR, @fdemir. I have to consider how we simulate a signed-in user for the tests. Please allow me some time to devise a solution to unblock this PR.

ycmjason:
@iuioiua hey i have upgraded my own project to the latest `kv_oauth` and have successfully simulated user sign in. Here's how I've done it.

```ts
async function fakeSignIn(user: User): Promise<void> {
  await createUser(user);
  await kv.set(["site_sessions", user.sessionId], true);
}
```

`kv_oauth` can't pick up the data from that key because each instances of sqlite opened using `:memory:` have data seperated. After looking into [how sqlite work](https://www.sqlite.org/inmemorydb.html) as guided from the [deno doc](https://docs.deno.com/deploy/kv/manual/#:~:text=Just%20like%20in%2Dmemory%20SQLite%20databases%2C%20multiple%20ephemeral%20Deno%20KV%20stores%20can%20exist%20at%20once%20without%20interfering%20with%20one%20another.%20For%20more%20information%20about%20special%20database%20addressing%20modes%2C%20see%20the%20SQLite%20docs%20on%20the%20topic.), we just have to change the database filename from `:memory:` to `file::memory:?cache=shared`.

so my test script become:

```
    "test": "DENO_KV_PATH=\"file::memory:?cache=shared\" deno test -A --parallel --unstable-kv",
```

Voila, this works like a charm.

Hope this helps!

Amazing project btw, saved me so much time. ❤️

iuioiua:
Hm... That seems quite hacky. Good find, nevertheless. Let me get back to you on what we do here.

iuioiua:
I find it hacky because it uses a barely-known KV path. I don't currently recall what the blocker was here. I'll look into this again once things calm down with `std` and the upcoming Deno 2 release.

PR #678, opened 2024-05-19
fix: correct typo on the pricing demo page

cassidoo:
suites -> suits


iuioiua (approved):
Good catch. Thank you.

PR #679, opened 2024-06-03
fix: added hotfix for submit page

harshmangalam:
- fix url label
- added url input id
- removed empty `<p>` tag
- fix welcome page unused vars


iuioiua (approved):
Thank you for this!

Issue #689, opened 2024-10-16
Bug report


iojanis:
Followed the guide til the first execution after setting Github keys:

```
deno task start
Task start deno run --unstable-kv -A --watch=static/,routes/ --env dev.ts
Watcher Process started.
error: Import assertions are deprecated. Use `with` keyword, instead of 'assert' keyword.

import emojis from "./all.json" assert { type: "json" };

  at https://deno.land/x/emoji@0.3.0/emoji.ts:2:1

Watcher Process failed. Restarting on file change...
´´´

**Comment by pchinjr** [comment]
I don't know where the application calls the emoji package, however on Deno versions 2.0.0 and above this error blocks the watcher and the server never starts. On Deno versions below 2.0.0 the error is displayed, but the server will continue to start and the application will run locally. 

```
pchinjr@WSL2:~/Code/saaskit$ deno upgrade --version 1.46.3
Current Deno version: v1.46.2
Downloading https://github.com/denoland/deno/releases/download/v1.46.3/deno-x86_64-unknown-linux-gnu.zip
Deno is upgrading to version 1.46.3

Upgraded successfully to Deno v1.46.3 (stable)

pchinjr@WSL2:~/Code/saaskit$ deno task start
Task start deno run --unstable-kv -A --watch=static/,routes/ --env dev.ts
Watcher Process started.
⚠️  Import assertions are deprecated. Use `with` keyword, instead of 'assert' keyword.

import emojis from "./all.json" assert { type: "json" };

  at https://deno.land/x/emoji@0.3.0/emoji.ts:2:1

The manifest has been generated for 22 routes and 3 islands.

 🍋 Fresh ready 
    Local: http://localhost:8000/

GA4_MEASUREMENT_ID environment variable not set. Google Analytics reporting disabled.
```
And with Deno version 2.0
```
pchinjr@WSL2:~/Code/saaskit$ deno upgrade --version 2.0.0
Current Deno version: v1.46.3
Downloading https://github.com/denoland/deno/releases/download/v2.0.0/deno-x86_64-unknown-linux-gnu.zip
Deno is upgrading to version 2.0.0

Upgraded successfully to Deno v2.0.0 (stable)

Migration guide:

  https://docs.deno.com/runtime/manual/advanced/migrate_deprecations

Release notes:

  https://github.com/denoland/deno/releases/tag/v2.0.0

Blog post:

  https://deno.com/blog/v2.0

pchinjr@WSL2:~/Code/saaskit$ deno task start
[...]

PR #692, opened 2024-10-24
Fix for issue #689: upgrade the routes/blog deno/gfm import

dominikj111:
`deno.land/x` is not on latest track of that package (see the https://github.com/denoland/deno-gfm/issues/131#issuecomment-2435930773).

Therefore I replaced by `jsr:@deno/gfm` instead. `deno task start` is not blocked anymore.


wesenbergg (approved):

dominikj111:
I'll update the PR tomorrow

PR #706, opened 2025-04-22
refactor: re-write `timeAgo()` to use `Intl.DurationFormat`

iuioiua:
Inspired by https://github.com/jsr-io/jsr/pull/974.

PTAL @kt3k


timreichen:
It should also be possible to pass a locale object. We might need to do custom localization for `just now` and `ago`.

PR #713, opened 2025-06-13
chore: remove `cov:gen` task

iuioiua:
This task is no longer needed as the `--coverage` flag in the `deno test` automatically produces the lcov file.


kt3k (approved):
LGTM

PR #714, opened 2025-06-20
refactor: move welcome plugin into middleware

iuioiua:
In hindsight, having this middleware in a plugin makes things more difficult to understand and doesn't provide much value. This change also works towards a Fresh 2 migration.


kt3k (approved):
LGTM

PR #715, opened 2025-06-20
refactor: move blog plugin directly into routes

iuioiua:
In the same vein as #714.


kt3k:
can you resolve the conflict?

